PR #3, opened 2017-04-04
Basic model structure

steventux:
The basic data models for link checking.

### DIFF
--- a/db/migrate/20170403153216_create_links.rb
+++ b/db/migrate/20170403153216_create_links.rb
@@ -1,0 +1,9 @@
+class CreateLinks < ActiveRecord::Migration[5.0]
+  def change
+    create_table :links do |t|
+      t.string :uri
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20170403155048_create_jobs.rb
+++ b/db/migrate/20170403155048_create_jobs.rb
@@ -1,0 +1,9 @@
+class CreateJobs < ActiveRecord::Migration[5.0]
+  def change
+    create_table :jobs do |t|
+      t.datetime :completed_at
+
+      t.timestamps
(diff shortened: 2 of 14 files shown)


db/migrate/20170403153216_create_links.rb
```diff
@@ -0,0 +1,9 @@
+class CreateLinks < ActiveRecord::Migration[5.0]
+  def change
+    create_table :links do |t|
+      t.string :url
```
kevindew: being pedantic this should probably be uri

db/migrate/20170403155048_create_jobs.rb
```diff
@@ -0,0 +1,9 @@
+class CreateJobs < ActiveRecord::Migration[5.0]
+  def change
+    create_table :jobs do |t|
+      t.datetime :last_checked_at
```
kevindew: I think a job will only run once so we probably just need a completed_at timestamp here

db/migrate/20170403153216_create_links.rb
steventux: :+1: may as well get it right now.

db/migrate/20170403153216_create_links.rb
```diff
@@ -0,0 +1,9 @@
+class CreateLinks < ActiveRecord::Migration[5.0]
+  def change
+    create_table :links do |t|
+      t.string :uri
```
steventux: @kevindew I'm not sure if this restricts the length of the uri to a db-specific limit. Would `t.text` be better?

kevindew: I don't think it does because Postgres varchar doesn't have a limit like MySQL does

PR #5, opened 2017-04-04
Initial link checks

thomasleese:
Paired with @emmabeynon on this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,8 @@ gem "plek", "~> 1.12"
 gem "airbrake", "~> 5.4"
 gem "pg"
 
+gem "faraday", "~> 0.11"
+
 group :development, :test do
   gem "pry"
   gem "simplecov-rcov", "~> 0.2", require: false
@@ -18,7 +20,7 @@ group :development, :test do
   gem "sqlite3" # Remove this when you choose a production database
   gem "factory_girl_rails", "~> 4.7"
   gem "timecop"
-  gem "webmock", require: false
+  gem "webmock"
   gem "rspec-rails", "~> 3.4"
   gem "byebug" # Comes standard with Rails
 end
--- a/lib/uri_checker.rb
+++ b/lib/uri_checker.rb
@@ -15,7 +15,7 @@ module UriChecker
     end
 
     def has_errors?
-      !errors.empty?
+      errors.any?
     end
   end
 
@@ -51,11 +51,11 @@ module UriChecker
   end
 
   class HttpChecker
-    INVALID_TOP_LEVEL_DOMAINS = %w(xxx adult)
-    REDIRECT_LIMIT = 10
+    INVALID_TOP_LEVEL_DOMAINS = %w(xxx adult dating porn sex sexy singles)
+    REDIRECT_LIMIT = 8
     REDIRECT_WARNING = 2
-    TIMEOUT = 5
-    SLOW_RESPONSE_TIME = 2
+    RESPONSE_TIME_LIMIT = 15
+    RESPONSE_TIME_WARNING = 2.5
 
     attr_reader :uri, :redirect_history, :report
 
@@ -110,7 +110,7 @@ module UriChecker
       end_time = Time.now
       response_time = end_time - start_time
 
-      report.warnings[:slow_response] << "Slow response time" if response_time > SLOW_RESPONSE_TIME
+      report.warnings[:slow_response] << "Slow response time" if response_time > RESPONSE_TIME_WARNING
 
       return response if report.has_errors?
 
@@ -186,8 +186,8 @@ module UriChecker
 
     def run_connection_request(method)
       connection.run_request(method, uri, nil, nil) do |request|
-        request.options[:timeout] = TIMEOUT
-        request.options[:open_timeout] = TIMEOUT
+        request.options[:timeout] = RESPONSE_TIME_LIMIT
+        request.options[:open_timeout] = RESPONSE_TIME_LIMIT
       end
     end
 
--- a/lib/uri_checker.rb
+++ b/lib/uri_checker.rb
@@ -1,0 +1,213 @@
+require "uri"
+
+module UriChecker
+  class Report
+    attr_reader :errors, :warnings
+
+    def initialize(errors: nil, warnings: nil)
+      @errors = errors || Hash.new { |hash, key| hash[key] = [] }
+      @warnings = warnings || Hash.new { |hash, key| hash[key] = [] }
+    end
+
+    def merge!(other)
+      errors.merge!(other.errors) { |key, oldval, newval| oldval | newval }
+      warnings.merge!(other.warnings) { |key, oldval, newval| oldval | newval }
+    end
+
+    def has_errors?
+      !errors.empty?
+    end
+  end
+
+  class ValidUri
+    HTTP_URI_SCHEMES = %w(http https)
+    FILE_URI_SCHEMES = %w(file)
+
+    attr_reader :report, :options
+
+    def initialize(options = {})
+      @report = Report.new
+      @options = options
+    end
+
+    def call(uri)
+      parsed_uri = URI.parse(uri)
+
+      if parsed_uri.scheme.nil?
+        report.warnings[:no_scheme] << "No scheme given."
+      elsif HTTP_URI_SCHEMES.include?(parsed_uri.scheme)
+        report.merge!(HttpChecker.new(parsed_uri, options).call)
+      elsif FILE_URI_SCHEMES.include?(parsed_uri.scheme)
+        report.merge!(FileChecker.new(parsed_uri, options).call)
+      else
+        report.warnings[:unsupported_scheme] << "Unsupported scheme."
+      end
+
+      report
+    rescue URI::InvalidURIError
+      report.errors[:uri_invalid] << "Invalid URI"
+      report
+    end
+  end
+
+  class HttpChecker
+    INVALID_TOP_LEVEL_DOMAINS = %w(xxx adult)
+    REDIRECT_LIMIT = 10
+    REDIRECT_WARNING = 2
+    TIMEOUT = 5
+    SLOW_RESPONSE_TIME = 2
+
+    attr_reader :uri, :redirect_history, :report
+
+     def initialize(uri, redirect_history: [])
+      @uri = uri
+      @redirect_history = redirect_history
+      @report = Report.new
+    end
+
+    def call
+      check_redirects
+      return report if report.has_errors?
+
+      check_top_level_domain
+      check_credentials
+
+      head_response = check_head_request
+      return report if report.has_errors?
+
+      check_get_request if head_response && head_response.headers["Content-Type"] == "text/html"
+      return report if report.has_errors?
+
+      check_google_safebrowsing
+
+      report
+    end
+
+  private
+
+    def check_redirects
+      report.errors[:too_many_redirects] << "Too many redirects." if redirect_history.length >= REDIRECT_LIMIT
+      report.errors[:cyclic_redirects] << "Has a cyclic redirect." if redirect_history.include?(uri)
+      report.warnings[:multiple_redirects] << "Multiple redirects." if redirect_history.length == REDIRECT_WARNING
+    end
+
+    def check_top_level_domain
+      tld = uri.host.split(".").last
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -18,7 +20,7 @@ group :development, :test do
   gem "sqlite3" # Remove this when you choose a production database
   gem "factory_girl_rails", "~> 4.7"
   gem "timecop"
-  gem "webmock", require: false
+  gem "webmock"
```
kevindew: do we need this change? it looks like webmock is required in the `spec/spec_helper.rb` which should do it? - unless I'm missing something different

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      @warnings = warnings || Hash.new { |hash, key| hash[key] = [] }
+    end
+
+    def merge!(other)
+      errors.merge!(other.errors) { |key, oldval, newval| oldval | newval }
+      warnings.merge!(other.warnings) { |key, oldval, newval| oldval | newval }
+    end
+
+    def has_errors?
+      !errors.empty?
```
steventux: Could be simplified to `errors.any?`

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
+require "uri"
```
kevindew: is this not required by default?

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      warnings.merge!(other.warnings) { |key, oldval, newval| oldval | newval }
+    end
+
+    def has_errors?
+      !errors.empty?
+    end
+  end
+
+  class ValidUri
+    HTTP_URI_SCHEMES = %w(http https)
```
steventux: Not a concern for this PR but would using a combination of `config_for` and `.yml` files help managing the many config options in these Checker classes?

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
+require "uri"
+
+module UriChecker
```
kevindew: I think it'd make sense that the link checking classes follow a namespace (or series of them) so I'd guess at LinkCheck::UriChecker.

I think they'd also be better separated out to multiple files as this has become quite a long file and could be confusing to find the classes

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+  class Report
```
kevindew: might want to add an `add_error` and `add_warning` method to this to save manipulating the attributes each time it's used

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+    attr_reader :errors, :warnings
+
+    def initialize(errors: nil, warnings: nil)
+      @errors = errors || Hash.new { |hash, key| hash[key] = [] }
+      @warnings = warnings || Hash.new { |hash, key| hash[key] = [] }
+    end
+
+    def merge!(other)
```
kevindew: linter might flag you on this, but you should try avoid using bang methods when there isn't a non bang alternative.

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      check_top_level_domain
+      check_credentials
+
+      head_response = check_head_request
+      return report if report.has_errors?
+
+      check_get_request if head_response && head_response.headers["Content-Type"] == "text/html"
+      return report if report.has_errors?
+
+      check_google_safebrowsing
```
kevindew: does this work ok for non text/html mime types?

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+
+     def initialize(uri, redirect_history: [])
+      @uri = uri
+      @redirect_history = redirect_history
+      @report = Report.new
+    end
+
+    def call
+      check_redirects
+      return report if report.has_errors?
```
kevindew: this seems premature as you could have a redirect with the below problems (unless I'm getting something wrong)

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+    def check_redirects
+      report.errors[:too_many_redirects] << "Too many redirects." if redirect_history.length >= REDIRECT_LIMIT
+      report.errors[:cyclic_redirects] << "Has a cyclic redirect." if redirect_history.include?(uri)
+      report.warnings[:multiple_redirects] << "Multiple redirects." if redirect_history.length == REDIRECT_WARNING
+    end
+
+    def check_top_level_domain
+      tld = uri.host.split(".").last
+      if INVALID_TOP_LEVEL_DOMAINS.include?(tld)
+        report.warnings[:risky_tld] << "Potentially suspicious top level domain."
```
kevindew: might want to include the tld in this

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      response = make_request(:head)
+      end_time = Time.now
+      response_time = end_time - start_time
+
+      report.warnings[:slow_response] << "Slow response time" if response_time > SLOW_RESPONSE_TIME
+
+      return response if report.has_errors?
+
+      if response.status >= 400 && response.status < 500
+        report.errors[:http_client_error] << "Received 4xx response"
```
kevindew: Be good to have actual status code here

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      elsif response.status >= 500 && response.status < 600
+        report.errors[:http_server_error] << "Received 5xx response"
```
kevindew: ditto previous

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      end
+
+      report.warnings[:http_non_200] << "Page not available." unless response.status == 200
```
kevindew: We probably want a different warning message here, but I'm not really sure what. It's kind of along the lines of: This link is returning an unexpected status code (150) are you sure it's working correctly? - it's kind of an annoying one as I imagine the likelihood of someone getting this is very low

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+            platformTypes: %w(ANY_PLATFORM),
+            threatEntryTypes: %w(URL),
+            threatEntries: [{ url: uri.to_s }]
+          }
+        }.to_json
+      end
+
+      data = JSON.parse(response.body)
+      if data.include?("matches") && data["matches"]
+        report.warnings[:google_safebrowsing] << "Google Safebrowsing has detected a threat."
```
kevindew: probably good to specify the sort of threat

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+            clientId: "gds-link-checker", clientVersion: "0.1.0"
+          },
+          threatInfo: {
+            threatTypes: %w(THREAT_TYPE_UNSPECIFIED MALWARE SOCIAL_ENGINEERING UNWANTED_SOFTWARE POTENTIALLY_HARMFUL_APPLICATION),
+            platformTypes: %w(ANY_PLATFORM),
+            threatEntryTypes: %w(URL),
+            threatEntries: [{ url: uri.to_s }]
+          }
+        }.to_json
+      end
```
kevindew: what happens if this request fails?

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+    end
+
+    def check_head_request
+      start_time = Time.now
+      response = make_request(:head)
+      end_time = Time.now
+      response_time = end_time - start_time
+
+      report.warnings[:slow_response] << "Slow response time" if response_time > RESPONSE_TIME_WARNING
```
steventux: Possibly not an immediate concern but I wonder if hard-wiring the error and warning messages should get refactored? It feels a bit inflexible. What's the value in storing a symbol and string message here?

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+      if data.include?("matches") && data["matches"]
+        report.warnings[:google_safebrowsing] << "Google Safebrowsing has detected a threat."
+      end
+    end
+
+    def make_request(method)
+      begin
+        response = run_connection_request(method)
+
+        if response.status >= 300 && response.status < 400
```
kevindew: We might want to limit the range of redirects we check or check for presence of location header. I know there's some weird ones: http://www.restapitutorial.com/httpstatuscodes.html and then there's also ones we shouldn't get like 304 but would cause us pain if we did get them.

lib/uri_checker.rb
thomasleese: Yeah, I think this is probably a good idea.

lib/uri_checker.rb
```diff
@@ -0,0 +1,213 @@
...
+
+  class FileChecker
+    attr_reader :report
+
+    def initialize(uri, options = {})
+      @report = Report.new
+    end
+
+    def call
+      report.errors[:local_file] << "Link is to a local file."
```
kevindew: might want to refactor this message to be a file not available on the internet (possibly your local computer)

lib/uri_checker.rb
thomasleese: I think we'll get value if we choose to include information in the string specific to the current URI which might change as we follow the redirects.

Gemfile
kevindew: ah just saw the last file - where that is changed to "webmock/rspec" any reason why we've done that? I don't know pros and cons just noticed it's different to Pub API

spec/lib/link_check_spec.rb
```diff
@@ -0,0 +1,228 @@
+require "uri_checker"
```
kevindew: If you put `lib/` in the eager load path you'll not need to require these explicitly (which is also a problem you'll have in the app.

lib/uri_checker.rb
steventux: Hadn't thought of this :+1: that sounds like a very useful case in point.

Gemfile
thomasleese: For some reason we had to do that to make it run the `WebMock.disable_net_connect!(allow_localhost: true)` line properly.

lib/uri_checker.rb
thomasleese: Hmm, I always thought of bang methods as when they are modifying the class itself.

lib/uri_checker.rb
thomasleese: Yes, we only provide the Google Safebrowsing API with a URL. It is able to check URLs that are links to programs, web pages, many other things.

lib/uri_checker.rb
thomasleese: We've sent Ben the content designer off to write all these errors for us. 😄

lib/uri_checker.rb
thomasleese: We've sent Ben the content designer off to write all these errors for us. 😄

lib/uri_checker.rb
thomasleese: We've sent Ben the content designer off to write all these errors for us. 😄 

Not sure he thinks it's a good idea to include the status code here as that's quite technical.

lib/uri_checker.rb
thomasleese: We've sent Ben the content designer off to write all these errors for us. 😄

lib/uri_checker.rb
thomasleese: It is!

lib/uri_checker.rb
kevindew: they're unfortunately quite inconsistent, so while you're understanding is correct in some circumstances it's not always the case. The most consistent definition is that they offer a dangerous alternative of a method. 

So `downcase` (a lovely translation oddity) is one where you've got `downcase` non modifying and `downcase!` modifying

whereas if you take ActiveRecord `save` and `save!` the bang indicates that it'll throw an exception. 

Since the language is object oriented there are lots of places where state is changed by a method call too so can quickly become a bit of a nightmare applying the bang modify pattern, whereas it gets a lots simpler applied to just dangerous alternative

lib/uri_checker.rb
kevindew: yeah sure. I'm expecting we're probably want specific ones for the biggies of 404 and maybe 401/403 and then a generic for the others

lib/uri_checker.rb
thomasleese: We will Airbrake it.

PR #7, opened 2017-04-04
Associate checks to a link

steventux:
Checks are specific to a single link. I've deliberately stuck to a simple association here fully expecting it to get more complex when necessary.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -22,6 +22,8 @@ ActiveRecord::Schema.define(version: 20170404125450) do
     t.json     "link_errors"
     t.datetime "created_at",    null: false
     t.datetime "updated_at",    null: false
+    t.integer  "link_id"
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -22,6 +22,8 @@
     t.json     "link_errors"
     t.datetime "created_at",    null: false
     t.datetime "updated_at",    null: false
+    t.integer  "link_id"
+    t.index ["link_id"], name: "index_checks_on_link_id", using: :btree
```
thomasleese: How come the date of the `schema.rb` hasn't change here?

steventux: Ah because I started adding the migration here before I realised the collision with AR `errors`, so the migration here is older than the one in https://github.com/alphagov/link-checker/pull/6 which shouldn't matter once both are merged should it? I've confused myself now.

PR #8, opened 2017-04-04
[Do Not Merge] Initial /check-link request spec

kevindew:
This is to try define the HTTP interface somewhat, may change as we
learn more.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  get "/check-link", to: proc { [200, {}, [{}.to_json]] }
 end
--- a/spec/requests/check_link_spec.rb
+++ b/spec/requests/check_link_spec.rb
@@ -1,0 +1,115 @@
+require "rails_helper"
+
+RSpec.describe "check link path", type: :request do
+  let(:link) { "http://www.example.com" }
+
+  def check_link_path(query_params = {})
+    path = "/check-link"
+    path + (query_params.empty? ? "?#{query_params.to_query}" : "")
+  end
+
+  def build_link_report(params)
+    {
+      "uri"       => params.fetch(:uri, anything),
+      "status"    => params.fetch(:status, anything),
+      "checked"   => params.fetch(:checked, anything),
+      "errors"    => params.fetch(:errors, []),
+      "warnings"  => params.fetch(:warnigns, []),
+    }
+  end
+
+  shared_examples "returns link report" do
+    it "returns 200" do
+      expect(response).to have_http_status(200)
+    end
+
+    it "returns a link report" do
+      json = JSON.parse(response.body)
+      expect(json).to match(link_report)
+    end
+  end
+
+  context "when no uri is requested" do
+    before do
+      get "/check-link"
+    end
+
+    it "returns 400" do
+      expect(response).to have_http_status(400)
+    end
+  end
+
+  context "when an ok checked uri is requested" do
+    let(:link_report) { build_link_report(uri: link, status: "ok") }
+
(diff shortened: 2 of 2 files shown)


config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  get "/check-link", to: proc { [200, {}, [{}.to_json]] }
```
thomasleese: I'm not sure about the need for `-link` here as that's fairly obvious from the service. Unless the plan for the batch endpoint is to be something like `/check-batch`?

spec/requests/check_link_spec.rb
```diff
@@ -0,0 +1,115 @@
+require "rails_helper"
+
+RSpec.describe "check link path", type: :request do
+  let(:link) { "http://www.example.com" }
+
+  def check_link_path(query_params = {})
+    path = "/check-link"
+    path + (query_params.empty? ? "?#{query_params.to_query}" : "")
```
thomasleese: I think `"/check-link" + ...` would look fine here, no need for the extra variable.

spec/requests/check_link_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    before { get check_link_path(uri: link) }
+
+    include_examples "returns link report"
+  end
+
+  context "when a checked uri, that is of status caution, is requested" do
+    let(:warnings) do
+      {
+        "risky_tld" => ["Potentially suspicious top level domain."],
```
thomasleese: I don't know if we should be testing the exact contents of the human readable strings as they might want to change reasonably frequently, whereas the error type slug will never change.

config/routes.rb
kevindew: Yeah, that's a good point, I'm not sure on that either - calling it just `/check` does limit the scope of it given it's a "checker" which could be argued to be either a good or bad thing. Also unsure if it should be "link" or 'uri" I suppose since we have a query string of `uri=` it does make it more redundant

spec/requests/check_link_spec.rb
kevindew: Agree - it was once most complex when we were interpolating a uri

spec/requests/check_link_spec.rb
kevindew: These wouldn't be exact ones (despite me c+p'ing an exact one, which will be the root of the confusion) it'd be whatever we defined in the warnings for the factory girl check instance

config/routes.rb
thomasleese: I think we should stick to `uri` as much as possible internally, but we can expose them to the end user as links. At the very least, we save one byte for each check using `uri` over `link`. 😉

spec/requests/check_link_spec.rb
thomasleese: Makes sense!

spec/requests/check_link_spec.rb
thomasleese: `(query_params.empty? ? "" : "?#{query_params.to_query}")` might work better 😆

spec/requests/check_link_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    path + (query_params.empty? ? "?#{query_params.to_query}" : "")
+  end
+
+  def build_link_report(params)
+    {
+      "uri"       => params.fetch(:uri, anything),
+      "status"    => params.fetch(:status, anything),
+      "checked"   => params.fetch(:checked, anything),
+      "errors"    => params.fetch(:errors, []),
+      "warnings"  => params.fetch(:warnigns, []),
```
emmabeynon: Just a small typo here `:warningns`

thomasleese:
Implemented and cherry picked in #11.

PR #9, opened 2017-04-04
Tidy up the link checking code

thomasleese:
Following on from @kevindew's review of #5.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,6 +22,9 @@ module LinkChecker
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    config.eager_load_paths << "#{config.root}/lib"
+    config.eager_load_paths += Dir["#{config.root}/lib/**/"]
+
     config.api_only = true
 
     # GDS SSO requires a session to exist
--- a/lib/link_check.rb
+++ b/lib/link_check.rb
@@ -7,7 +7,6 @@ class LinkCheck
   end
 
   def call
-    report.merge!(UriChecker::ValidUri.new.call(uri))
-    report
+    report.merge(UriChecker::ValidUri.new.call(uri))
   end
 end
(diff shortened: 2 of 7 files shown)


config/application.rb
```diff
@@ -22,6 +22,10 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    config.eager_load_paths << "#{config.root}/app"
```
kevindew: don't think we need this - probably don't want to encourage classes in root of app.

lib/link_check.rb
```diff
@@ -3,11 +3,10 @@ class LinkCheck
 
   def initialize(uri)
     @uri = uri
-    @report = UriChecker::Report.new
+    @report = LinkCheck::UriChecker::Report.new
```
kevindew: these should still work without the namespace you've added in as it'll be scoped to `LinkCheck` already

lib/link_check/uri_checker/http_checker.rb
```diff
@@ -1,55 +1,4 @@
...
-      report
-    rescue URI::InvalidURIError
-      report.errors[:uri_invalid] << "Invalid URI"
-      report
-    end
-  end
-
+module LinkCheck::UriChecker
   class HttpChecker
     INVALID_TOP_LEVEL_DOMAINS = %w(xxx adult dating porn sex sexy singles)
```
kevindew: we should host this on one of those. `link-checker.sexy`

lib/link_check/uri_checker/http_checker.rb
```diff
@@ -154,33 +101,41 @@ def check_google_safebrowsing
...
+        })
       end
     end
 
     def make_request(method)
       begin
         response = run_connection_request(method)
 
-        if response.status >= 300 && response.status < 400
+        if [301, 302, 303, 307, 308].include?(response.status) && !report.has_errors?
```
kevindew: probably want to check for location header in this conditional too

lib/link_check/uri_checker/http_checker.rb
thomasleese: `link-checker.singles`

PR #10, opened 2017-04-05
[Do Not Merge] Basic /batch endpoint specs

kevindew:
An attempt to define the interface and specs for the /batch endpoint.

Marked as DNM due to it failing tests

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  post "/batch", to: proc { [202, {}, [{}.to_json]] }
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  post "/batch", to: proc { [202, {}, [{}.to_json]] }
+  get "/batch/:id", to: proc { [200, {}, [{}.to_json]] }
```
thomasleese: If `/check` and `/batch` both create jobs (for example when being asynchronous), I think it would be better if the GET variation is `GET /job/:id` since really we're getting information about a job.

thomasleese:
Implemented and cherry picked in #11.

PR #11, opened 2017-04-05
Jobs and Controllers

thomasleese:
Implementation of the various jobs and controllers of the link checker.


app/models/check.rb
```diff
@@ -1,3 +1,36 @@
...
+    link_warnings.any?
+  end
+
+  def is_ok?
+    !has_errors? && !has_warnings?
+  end
+
+  def status
+    return "pending" if is_pending?
+    return "error" if has_errors?
```
kevindew: we renamed this "broken" (I think that was agreed?)

app/models/link.rb
```diff
@@ -1,4 +1,18 @@
...
+    checks
+      .where(ended_at.gt(24.hours.ago))
+      .first
+  end
+
+private
+
+  def ended_at
+    Check.arel_table[:ended_at]
+  end
```
kevindew: what's this?

kevindew: Oh I see

app/models/link.rb
```diff
@@ -1,4 +1,18 @@
...
   has_and_belongs_to_many :jobs
+  has_many :checks
+
   validates_presence_of :uri
+
+  def existing_check
+    checks
+      .where(ended_at.gt(24.hours.ago))
+      .first
+  end
```
kevindew: I imagine this will be configurable by input. Might be more idiomatic to do this with `("ended_at > ?", Time.now - input)`

kevindew: Also in a different area of concern, I wonder if we want ended_at to be `completed_at` as I think that's what we have on a Job (or batch)

app/jobs/link_check_job.rb
```diff
@@ -0,0 +1,31 @@
+class LinkCheckJob < ApplicationJob
+  queue_as :default
+
+  # TODO:
+  #   - The Check#ended_at db query param should be stored in the job when
+  #     it's persisted.
+  #   - Creating the Check record could be more elegant.
+  def perform(job)
```
kevindew: should take in a link rather than a job here

thomasleese:
Things left to do:

- Investigate `govuk-sidekiq`
- Make the webhook job run immediately after a batch or a check finishes, rather than currently where it might be up to 5 seconds after.

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,17 @@
...
   before_action :require_signin_permission!
 
-  protect_from_forgery with: :exception
+  rescue_from ActionController::ParameterMissing do |e|
+    render json: { error: { message: e.message } }, status: 400
+  end
+
+  rescue_from ActiveRecord::RecordNotFound do |e|
+    render json: { error: { message: e.message } }, status: 404
+  end
```
kevindew: I think we get this for free and don't need to catch it

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,17 @@
...
+
+  def payload
+    @payload ||= JSON.parse(request.body.read).deep_symbolize_keys
+  end
```
kevindew: This looks familiar :wink: - I don't think the precedent set by Publishing API is a great one to follow though as it fights against rails conventions

if, on a post request, we have a header of content-type application/json Rails should automatically set the params to be json parsed, it should also be a [HashWithIndifferentAccess](http://api.rubyonrails.org/v4.2/classes/ActiveSupport/HashWithIndifferentAccess.html) so wouldn't need the deep symbolize either

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,37 @@
+class BatchController < ApplicationController
+  def create
+    uris = payload.fetch(:uris)
+    checked_within = (payload[:checked_within] || 24.hours).to_i
+    callback_uri = payload[:callback_uri]
```
kevindew: it'd be nice to validate these - rails doesn't make that super easy outside of db models. But a quick google around shows this might be pretty good: http://api.rubyonrails.org/classes/ActiveModel/Validations.html

failing that this is a less railsy alternative: http://dry-rb.org/gems/dry-validation/

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    checks = links.map do |link|
+      check = link.find_completed_check(within: checked_within)
+      check ? check : Check.create!(link: link)
+    end
+
+    batch = Batch.create!(checks: checks)
+
+    if batch.completed?
+      WebhookJob.perform_now(batch, callback_uri) if callback_uri
```
kevindew: I imagine we probably want to still run this asynchronously since the IO could hold up our response and we don't use any of the response from this to complete our response

app/controllers/check_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    check = Check.create!(link: link)
+
+    if synchronous
+      CheckJob.perform_now(check, callback_uri: callback_uri)
+    else
+      CheckJob.perform_later(check, callback_uri: callback_uri)
+    end
+
+    render(json: check.to_h)
+  end
```
kevindew: I'm not sure there's much of a use case for callbacks for just a single link, feels like you'd use the `/check` endpoint either for debugging or if you couldn't handle the batch/callbacks and reluctantly pushed us into synchronous mode

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    return render json: { error: { message: "No URIs given." } }, status: 400 if uris.empty?
+
+    links = uris.map do |uri|
+      Link.find_or_create_by(uri: uri)
+    end
+
+    checks = links.map do |link|
+      check = link.find_completed_check(within: checked_within)
+      check ? check : Check.create!(link: link)
+    end
```
kevindew: both of these loops feel like they'd make a bunch of queries when single ones could do most of it. 

I'd consider extracting it to a class - maybe `Batch::CreateChecks` and you could create a batch object first (I expect it'd be useful to store callback and checked_within values on it) which handles it? 

might also want to wrap the batch stuff in a transaction and see if we can do what we can to limit dangers of two requests at same time

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    callback_uri = payload[:callback_uri]
+
+    return render json: { error: { message: "No URIs given." } }, status: 400 if uris.empty?
+
+    links = uris.map do |uri|
+      Link.find_or_create_by(uri: uri)
+    end
+
+    checks = links.map do |link|
+      check = link.find_completed_check(within: checked_within)
```
kevindew: I imagine we want to have pending checks included here as well

app/jobs/check_job.rb
```diff
@@ -0,0 +1,25 @@
+class CheckJob < ApplicationJob
+  queue_as :default
+
+  def perform(check, batch: nil, callback_uri: nil)
```
kevindew: Rather than a single batch as an argument, it'd probably more efficient to have a query that pulls out all batches that are affected by a link

app/jobs/check_job.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    check.update!(
+      link_errors: report.errors,
+      link_warnings: report.warnings,
+      completed_at: Time.now
+    )
+
+    if callback_uri
+      if batch
+        WebhookJob.perform_now(batch, callback_uri) if batch.completed?
```
kevindew: I think we probably want to hit the callback each time there is progress, that'd allow apps to gradually update their ui rather than waiting for the whole thing?

app/jobs/webhook_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  queue_as :default
+
+  def perform(check_or_batch, callback_uri)
+    response = Faraday.post do |req|
+      req.url callback_uri
+      req.headers["Content-Type"] = "application/json"
+      req.body = check_or_batch.to_h.to_json
+    end
+
+    if response.status >= 500
```
kevindew: can we get an exception back from faraday when a request fails? it feels like it'd be better to use the [ActiveJob retry mechanism](http://api.rubyonrails.org/classes/ActiveJob/Enqueuing.html) - also worth checking that ActiveJob retry eventually aborts so we don't get into an eternal loop (which we'd fall into with this implementation if a server consistently 500'd)

And finally we probably want to be retrying on more cases than 500

app/models/batch.rb
```diff
@@ -0,0 +1,36 @@
...
+      totals: {
+        links: checks.count,
+        ok: checks.each.count { |check| check.status == "ok" },
+        caution: checks.each.count { |check| check.status == "caution" },
+        broken: checks.each.count { |check| check.status == "broken" },
+        pending: checks.each.count { |check| check.status == "pending" },
+      },
+      completed_at: completed_at,
+    }.deep_symbolize_keys
+  end
```
kevindew: I think this belongs in a presenter class, probably of name `BatchReport` so we can be consistent with terminology (hopefully we don't already have a collision on LinkReport 😬 )

I'd suggest that `to_h` is more of a place to serialize the model in the same way as `as_json` whereas we're presenting this as a domain concept.

app/models/batch.rb
```diff
@@ -0,0 +1,36 @@
...
+    checks.maximum(:completed_at) if completed?
+  end
+
+  def completed?
+    checks.where(completed_at: nil).empty?
+  end
+
+  def status
+    return "completed" if completed?
+    return "in_progress"
```
kevindew: `completed? ? "completed" : "in_progress"`

app/models/batch.rb
```diff
@@ -0,0 +1,36 @@
...
+    checks.minimum(:started_at)
+  end
+
+  def completed_at
+    checks.maximum(:completed_at) if completed?
+  end
+
+  def completed?
+    checks.where(completed_at: nil).empty?
+  end
```
kevindew: These methods might all be running queries each time they're used. Might want to iterate through checks to find the items which would save queries. I'm not sure on this though, would need some experimenting.

app/models/check.rb
```diff
@@ -1,3 +1,49 @@
...
+
+  def to_h
+    {
+      uri: link.uri,
+      status: status,
+      checked: completed_at,
+      errors: link_errors,
+      warnings: link_warnings,
+    }.deep_symbolize_keys
+  end
```
kevindew: same point on presenter as batch

app/models/check.rb
```diff
@@ -1,3 +1,49 @@
...
+
+  def completed?
+    !is_pending?
+  end
+
+  def status
+    @status ||= determine_status
+  end
+
+  after_save { remove_instance_variable(:@status) if defined? @status }
```
kevindew: does this need to be an ivar? be good if we could remove need for a callback.

app/models/check.rb
```diff
@@ -1,3 +1,49 @@
...
+  def has_errors?
+    link_errors.any?
+  end
+
+  def has_warnings?
+    link_warnings.any?
+  end
+
+  def is_ok?
+    !has_errors? && !has_warnings?
```
kevindew: could still be pending here? might be better to use `status == "ok"`

app/models/check.rb
```diff
@@ -1,3 +1,49 @@
...
+
+  after_save { remove_instance_variable(:@status) if defined? @status }
+
+private
+
+  def determine_status
+    return "pending" if is_pending?
+    return "broken" if has_errors?
+    return "caution" if has_warnings?
+    return "ok"
```
kevindew: the return is implicit 😉  also would probably feel more rubyish if we used symbols here, unless I'm missing some disadvantage?

app/models/link.rb
```diff
@@ -1,4 +1,11 @@
...
-  has_and_belongs_to_many :jobs
+  has_many :checks
+
   validates_presence_of :uri
+
+  def find_completed_check(within: 24.hours)
+    checks
+      .where("completed_at > ?", Time.now - within)
+      .first
+  end
```
kevindew: I think, but not sure, that this might be better as a scope on `Check` it'd help differentiate it as it executes a query

Should also have an ordering to get most recent one.

config/application.rb
```diff
@@ -16,7 +16,7 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-module LinkChecker
+module LinkCheckerApi
```
kevindew: 🎉

config/sidekiq.yml
```diff
@@ -0,0 +1,2 @@
+---
+:concurrency:  2
```
kevindew: Hopefully we can ramp this up a bit

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,17 @@
-class ApplicationController < ActionController::API
+class ApplicationController < ActionController::Base
```
thomasleese: The GDS SSO uses `helper_method` which is not available in the `API` controller.

app/controllers/application_controller.rb
thomasleese: If I don't put this in, the `ActiveRecord::RecordNotFound` error gets passed up all the way to the test, so it seems like we do?

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,17 @@
...
   include GDS::SSO::ControllerMethods
 
   before_action :require_signin_permission!
 
-  protect_from_forgery with: :exception
+  rescue_from ActionController::ParameterMissing do |e|
+    render json: { error: { message: e.message } }, status: 400
+  end
```
thomasleese: We still have to include the `rescue_from` method otherwise the `ParameterMissing` exception gets passed all the way to the test. Plus we get the features of strong parameters in Rails 5 now automatically, by using `fetch` we get that exact `ParameterMissing` exception.

app/controllers/application_controller.rb
kevindew: Yeah, I don't think it wraps it in test - I recall writing tests expecting that exception. Though try it on rails in dev and prod envs to see. 

There's a whole bunch of things wrapped around: https://github.com/rails/rails/blob/fe1f4b2ad56f010a4e9b93d547d63a15953d9dc2/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb

app/controllers/application_controller.rb
kevindew: That's a pain 👎  GDS SSO is a real thorn in rails API mode

app/controllers/application_controller.rb
thomasleese: Hmm, that feels wrong to me, how can we test that our error JSON blob matches what we expect if we just get an exception?

app/controllers/application_controller.rb
thomasleese: Yeah, I see we can use the `permit` feature, but I would just say we ignore extra parameters that we're given. I'm not passing the `params` hash around to anything directly, so there's no risk of accidental injection, but perhaps this is me fighting Rails.

app/models/check.rb
thomasleese: This is just an optimisation to avoid calculating redetermining the status if nothing about the model has changed. It doesn't need to be there, but it saves queries.

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      @params = params
+      @id = permitted_params[:id]
+    end
+
+    def permitted_params
+      @permitted_params ||= @params.permit(:id)
+    end
+  end
+
+  def show
```
kevindew: we get param as an argument here and won't need to validate it

app/models/batch.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def to_h
+    {
+      id: id,
+      status: status,
+      checks: checks.map(&:to_h),
+      started_at: started_at,
+      completed_at: completed_at,
+    }
+  end
```
kevindew: do we need this?

app/models/check.rb
```diff
@@ -1,3 +1,60 @@
 class Check < ApplicationRecord
+  has_and_belongs_to_many :batches
   belongs_to :link
+
+  scope :created_within, -> (within) { where("created_at > ?", Time.now - within) }
+
+  def self.fetch_all(links, within: 24.hours)
```
kevindew: maybe improve name? - not sure of ideas

app/presenters/batch_presenter.rb
```diff
@@ -0,0 +1,25 @@
+class BatchPresenter
```
kevindew: maybe try [SimpleDelegator](https://ruby-doc.org/stdlib-2.2.1/libdoc/delegate/rdoc/SimpleDelegator.html)

thomasleese:
Woo 🎉

PR #15, opened 2017-04-06
Add user-friendly responses

benjamin-mortimer:
I've added a short explanation to each response.

The existing responses are good labels and are very useful as they are if you know what's going on.

So my approach has been to leave them in place and add a short gloss for people who don't know what's going on, for example:

CHANGED "Unsupported scheme"
TO "Unsupported scheme - we can't check this type of link."

or

CHANGED  "Not a valid URI."
TO  "Not a valid URI or URL."
(No one seems to understand 'URI').

I've tried to be clever on lines 97 and 134 and use string interpolation. Shout if it's wrong/a bad idea.

Redirects could probably do with some more work.

### DIFF
--- a/lib/uri_checker.rb
+++ b/lib/uri_checker.rb
@@ -34,18 +34,18 @@ module UriChecker
       parsed_uri = URI.parse(uri)
 
       if parsed_uri.scheme.nil?
-        report.warnings[:no_scheme] << "No scheme given."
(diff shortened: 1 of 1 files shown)


benjamin-mortimer:
Haven't done a PR for...a while. Let me know if this is all wrong.

PR #36, opened 2017-04-10
Better retries

### LINKED ISSUE #32: Should we have a status for when we've failed to check a link
If Sidekiq exhausts all of it's retries on a link we're be left with our status as "pending" but we can catch this and set it to something. Not sure what to call it as we want to differentiate it from broken. Maybe "caution" and a warning that a problem with the link checker meant it couldn't be checked?

Probably also want to configure Sidekiq so it doesn't run for it's full 21 days before it does this.

thomasleese:
Closes #32.

### DIFF
--- a/app/workers/check_worker.rb
+++ b/app/workers/check_worker.rb
@@ -1,6 +1,8 @@
 class CheckWorker
   include Sidekiq::Worker
 
+  sidekiq_options retry: 3
+
   def perform(check_id)
     check = Check.includes(:link, :batches).find(check_id)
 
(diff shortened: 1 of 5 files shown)


app/workers/check_worker.rb
```diff
@@ -1,6 +1,21 @@
 class CheckWorker
   include Sidekiq::Worker
 
+  sidekiq_options retry: 10
```
kevindew: Do you know how long 10 retries takes until? I know 25 ~= 21 days so not sure if this is measured in days?

thomasleese:
```ruby
def backoff(retry_count)
  (retry_count ** 4) + 15 + (rand(30) * (retry_count + 1))
end

100.times.map { 10.times.map { |i| backoff(i) }.sum }.sum / 100
```

That gives me about 4.5 hours as how long it takes.

PR #41, opened 2017-04-10
Add Procfile to create Sidekiq process

emmabeynon:


Procfile
```diff
@@ -0,0 +1 @@
+worker: bundle exec sidekiq -C ./config/sidekiq.yml
```
thomasleese: I think we can include the `web` process here as well:

```
web: bundle exec rails s -p 3208
```

emmabeynon: Cool -wasn't sure if it was needed or not.  Will add it in 👍

PR #52, opened 2017-04-12
Conform to US spelling of License

kevindew:
We've ended up in the funny situation of having two licence files, it
probably makes sense to adopt the US spelling since:

- Most software licences originate in the US and adopt "License" in
  their title, thus a more common convention
- Github does magic things to a file called "LICENSE"


fofr:
* Other apps use LICENCE
* The copyright on the LICENSE that's been left is wrong. It should be "Copyright (C) YEAR Crown copyright (Government Digital Service)"

kevindew:
@fofr Thanks for that

The app consistency doesn't seem to be uniform [1], [2] - I think we could benefit from the rails app template to change.

[1]: https://github.com/alphagov/publishing-api/blob/master/LICENSE
[2]: https://github.com/alphagov/content-store/blob/master/LICENSE

PR #53, opened 2017-04-12
Secure Webhooks

thomasleese:
Also removes SSO.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,2 @@
 class ApplicationController < ActionController::Base
-  include GDS::SSO::ControllerMethods
-
-  before_action :require_signin_permission!
 end
--- a/app/workers/webhook_worker.rb
+++ b/app/workers/webhook_worker.rb
@@ -3,11 +3,16 @@ class WebhookWorker
 
   sidekiq_options queue: :webhooks
 
-  def perform(report, webhook_uri)
+  SIGNATURE_HEADER = "X-LinkCheckerApi-Signature".freeze
+
+  def perform(report, uri, secret_token)
+    body = report.to_json
+
     connection.post do |req|
-      req.url webhook_uri
+      req.url uri
       req.headers["Content-Type"] = "application/json"
-      req.body = report.to_json
+      req.headers[SIGNATURE_HEADER] = generate_signature(body, secret_token) if secret_token
+      req.body = body
     end
   end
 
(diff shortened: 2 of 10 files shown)


app/workers/webhook_worker.rb
```diff
@@ -17,4 +22,12 @@ def connection
       faraday.use Faraday::Response::RaiseError
     end
   end
+
+  def generate_signature(body, key)
+    OpenSSL::HMAC.hexdigest(digest, key, body)
```
kevindew: probably easier to read if we just put the digest directly in here

app/controllers/application_controller.rb
```diff
@@ -1,5 +0,0 @@
-class ApplicationController < ActionController::Base
```
kevindew: Even if it's empty we still probably want the ApplicationController class as it's a rails convention.

spec/workers/webhook_worker_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  describe "perform" do
+    let(:report) { { some: "json" } }
+    let(:webhook_uri) { "http://webhooks-rule.org/webhook" }
+    let(:secret_key) { nil }
+
+    context "with a secret key" do
+      let(:secret_key) { "this is a secret key" }
+
+      # http://www.freeformatter.com/hmac-generator.html
+      let(:expected_signature) { "d44d805a16b90922edb48f14ff4e858a5d3e39bd" }
```
kevindew: Since we can generate the signature with one line it might be better to just generate it here than using the external site to do it.

spec/workers/webhook_worker_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+                headers: { "X-LinkCheckerApi-Signature": expected_signature })
+          .to_return(status: 200)
+      end
+
+      before do
+        subject.perform(report, webhook_uri, secret_key)
+      end
+
+      it "generates a valid signature" do
+        expect(stubbed_request).to have_been_requested
```
kevindew: Might be more idiomatic to have this be along the lines of `expect(a_request(:post, webhook_uri).with(headers: {...})).to have_been_made` and then just do a more generic stub in the before block?

spec/workers/webhook_worker_spec.rb
thomasleese: Hmm, yes, I wasn't sure about this one. I went for the website approach so we can be 100% sure the HMAC is generated properly. I put the JSON in with the key and got the expected response.

If I just copy in the code from the signature generation method, then I would say we're not really testing that we get a valid HMAC situation, we're just running the same code twice.

thomasleese:
@kevindew Ready for re-review 😄

PR #56, opened 2017-04-12
Include Sidekiq Web UI in development

kevindew:


config/routes.rb
```diff
@@ -5,4 +5,9 @@
 
   post "/batch", to: "batch#create"
   get "/batch/:id", to: "batch#show"
+
+  if Rails.env.development?
+    require 'sidekiq/web'
```
thomasleese: Single quotes? 😉

kevindew: Drat foiled by a c+p! Thanks for catching

PR #61, opened 2017-04-13
Rerun checks on schedule

### LINKED ISSUE #57: Sidekiq is not resilient - work around it
I found from running the full Whitehall link checking process that I got caught with about 5 batches stuck in a perpetual `in_progress` state. Sidekiq job was empty, so I don't know what happened to mean they didn't get run. The cause here probably doesn't matter so much though, just background.

What's more important is that it is possible in our system - Sidekiq can crash and jobs could be lost - and we're stuck in eternal pending states.

There are probably a few ways to solve this, there's software available for it such as [attentive_sidekiq](https://github.com/twonegatives/attentive_sidekiq). Though I think our concerns are probably slightly deeper than a sidekiq level - as at the point a check is marked as pending it hasn't been pushed to sidekiq. The one that comes to my mind is to use a scheduled sidekiq job such as [sidekiq scheduler](https://github.com/moove-it/sidekiq-scheduler) and use that to check for pending checks that are kinda old (15/20 minutes ?) and requeue them?

thomasleese:
Closes #57.


app/workers/cleanup_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def perform
+    check_ids.each do |check_id|
+      CheckWorker.perform_async(check_id)
+    end
+  end
+
+  def check_ids
+    Check
+      .where(completed_at: nil)
```
kevindew: do we need this? figure it's probably covered by the scope?

Issue #54, opened 2017-04-12
Undefined method body for LinkChecker::UriChecker::Report


kevindew:
Just saw this error looks like it originates here: https://github.com/alphagov/link-checker-api/blob/master/lib/link_checker/uri_checker/http_checker.rb#L91 and is due to the link failing the request and a Report object being returned rather than a Response, code looks like it was anticipating a nil response in the case of a failure

thomasleese:
Good catch, need to return `nil` at the end of `make_request` since `add_error` and `add_warning` return `self` to enable chaining.

PR #62, opened 2017-04-13
Fix response for links failing GET request

emmabeynon:
Closes https://github.com/alphagov/link-checker-api/issues/54

### DIFF
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -146,12 +146,16 @@ module LinkChecker::UriChecker
         response
       rescue Faraday::ConnectionFailed
         report.add_error("Connection failed", "Connection to the server could not be established.")
+        nil
       rescue Faraday::TimeoutError
         report.add_error("Timeout error", "The connection to the server timed out.")
+        nil
       rescue Faraday::SSLError
         report.add_error("Unsafe link", "This site's SSL security certificate has expired - it might not be safe for users.")
+        nil
       rescue Faraday::Error => e
         report.add_error("Unknown issue", "Speak to your technical team: #{e}")
+        nil
       end
     end
 


spec/lib/link_checker_spec.rb
```diff
@@ -227,5 +227,16 @@
       include_examples "has a warning", "Flagged as dangerous"
       include_examples "has no errors"
     end
+
+    context "a uri fails the request" do
+      let(:uri) { "http://www.example.com/failed-response" }
+      before do
+        stub_request(:head, uri).to_return(status: 200)
+        stub_request(:post, uri)
```
thomasleese: I think this one needs to be `:get`.

emmabeynon: Yes! Will fix now.

kevindew:
I don't think this has covered issue #54 unfortunately - that probably means I gave you bad advice on how to replicate it. 

So the issue seems to be in this code: https://github.com/alphagov/link-checker-api/blob/master/lib/link_checker/uri_checker/http_checker.rb#L146-L154 where these errors return a `Report` object - whereas the function is expected to return either a `Response` object or a nil value.

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -133,6 +133,7 @@ def check_google_safebrowsing
     def make_request(method)
       begin
         response = run_connection_request(method)
+        return nil unless response
```
kevindew: this code will probably not get executed, as it would throw an exception and you'd get hoisted into the rescue blocks.

Side note: you also don't need to add nil for this scenario, an empty return implicitly returns nil

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -152,6 +152,7 @@ def make_request(method)
         report.add_error("Unsafe link", "This site's SSL security certificate has expired - it might not be safe for users.")
       rescue Faraday::Error => e
         report.add_error("Unknown issue", "Speak to your technical team: #{e}")
+        nil
```
kevindew: This looks the way to solve it, will need this for each of the rescues though

PR #68, opened 2017-06-30
Improved message format

thomasleese:
This PR implements [the proposed new message format](https://docs.google.com/a/digital.cabinet-office.gov.uk/document/d/13lT6g_l4EILDs2q2HjLoyvScukp4GjbXxPFRkDMaHlI/edit?usp=sharing) for checks.

[Trello Card](https://trello.com/c/NJqopHyX/514-implement-new-message-format)

### DIFF
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -53,4 +53,14 @@ class Check < ApplicationRecord
     return :caution if has_warnings?
     :ok
   end
+
+  def problem_summary
+    return nil if is_ok? || is_pending?
+    "A problem summary."
+  end
+
+  def suggested_fix
+    return nil if is_ok? || is_pending?
+    "A suggested fix."
+  end
 end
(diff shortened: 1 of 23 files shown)


kevindew:
Thanks for picking this up Tom 👍  - a few thoughts before a more thorough review. 

Couple of things I think could potentially be missed, or at least couldn't find explicit test for them.
- Only show a suggested_fix if there is one problem?
- Even if same problem occurs on multiple redirects only show the problem once?
- Did you look at the IRI handling?

I also wonder if things might be improved by:
- using a different way to order problems than priority (they're always tricky) - an idea would be that each grouping has a symbol that defines them and we just have an array somewhere that defines that ordering.
- Maybe we could move the copy of the messages somewhere else, maybe translations file? and just have a symbol in the code, to make that less verbose?

Also the sheet suggested we drop the warnings for things we don't support - they kept in for any particular reason?

thomasleese:
@kevindew I've addressed some of the points you've raised, regarding your other questions I think further discussion is necessary.

PR #75, opened 2017-07-11
Change default checked_within period

sihugh:
The `checked_within` period specifies how stale a URL check can be before the
consuming application considers it useful.  I consider the current default of
24 hours to be a bit long, and offer 4 hours as an alternative.

Local links manager (as one of the first consuming apps) needs to run a check
of all links on a nightly basis.  It is entirely possible that a check from
night to night will run a few seconds earlier and thus get no fresh check
results.

I'm aware that this value is overridable, but I guess that the use case of
doing a nightly run is likely to be a common one (or perhaps even the default),
so we should allow this to occur without friction for the user.

The other common use case is to do another run within the working day.  Local
government websites are sometimes off-line overnight for maintenance.  Doing
another run during the day so that we can get a better set of results
(especially if the site is slow to start after a service recycle or a period
of no requests) is also something that we'd want to do.  4 hours would allow this
 to occur twice within the work day without further configuration.

### DIFF
--- a/app/controllers/batch_controller.rb
+++ b/app/controllers/batch_controller.rb
@@ -11,7 +11,7 @@ class BatchController < ApplicationController
     def initialize(params)
       @params = params
       @uris = permitted_params[:uris]
-      @checked_within = (permitted_params[:checked_within] || 24.hours).to_i
+      @checked_within = (permitted_params[:checked_within] || 4.hours).to_i
       @priority = permitted_params.fetch(:priority, "high")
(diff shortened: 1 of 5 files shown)


kevindew:
I'm good with it 👍 

As LLM is our only client makes sense to get a good value for that. Can increase it again if we get problems with load but seems unlikely until more things use this.

sihugh:
Thanks both 👍

PR #79, opened 2017-07-14
Reduce queue ratio

thomasleese:
We have a problem at the moment where the webhook queue is
getting too big and filling up memory in Redis.


thomasleese:
We don't need to do this because of a better fix in https://github.com/alphagov/link-checker-api/pull/80.

PR #82, opened 2017-10-27
Correct development db name

SebAshton:
Data replication creates a db called link_checker_api_development not link_checker_development


kevindew:
Good catch, Do you want to squash these first? we don't need to merge the revert commit in

SebAshton:
@kevindew have squashed

PR #84, opened 2017-11-02
Add .editorconfig

SebAshton:
This helps me keep my editor in good shape :)

What do people think about adding it?

Find out more here: http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+tab_width = 2
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+tab_width = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
kevindew: ironically you should insert a trailing new line here 😆

SebAshton: 🤣  Have added that in 😄

PR #85, opened 2017-11-02
Implement HistoricalLinkChecker

SebAshton:
`HistoricalLinkChecker` wraps `LinkChecker` to look back on previous check results so that we can raise a different set of errors when a particular error has been recurring for a while.

## Why?

This will enable us to escalate and prioritise issues that have occurred for a longer period of time.

For example, in future, we can prioritise 500 or 404 errors that have been occurring for more than a week over ones that have only been happening for a day. Sometimes URLs will disappear temporarily when a website is being deployed, or error temporarily when a developer deploys a bug (whoops).

### DIFF
--- a/lib/historical_link_checker.rb
+++ b/lib/historical_link_checker.rb
@@ -1,0 +1,77 @@
+class HistoricalLinkChecker
+  attr_reader :report
+
+  def initialize(uri, checks)
+    @uri = uri
+    @checks = sorted_checks(checks)
+    @report = LinkChecker::UriChecker::Report.new
+  end
+
+  def call
+    current_report = LinkChecker.new(uri).call
+
+    if needs_historical_data_check?(current_report)
+      check_history(current_report)
+    else
+      current_report
+    end
+  end
+
+private
+
+  attr_reader :uri, :checks
+
+  def check_history(current_report)
+    if all_checks_have_errors?(checks_since(7.days.ago))
+      current_report.add_problem(
+        recurred_for_more_than_one_week_error(current_report)
+      )
+    elsif all_checks_have_errors?(checks_since(4.days.ago))
+      current_report.add_problem(
+        recurred_for_more_than_three_days_error(current_report)
+      )
+    else
+      current_report
+    end
+  end
+
+  def needs_historical_data_check?(current_report)
+    return false unless current_report.has_errors?
+    return false if checks.empty?
+    return false unless checks_have_errors?(checks)
+
+    true
+  end
+
+  def locale_interpolation_args(current_report)
+    { problem: current_report.problem_summary, problem_message: current_report.errors.try(:first) }
+  end
+
+  def recurred_for_more_than_one_week_error(current_report)
+    HistoricalLinkChecker::Error::RecurredForMoreThanOneWeekLinkError.new(locale_interpolation_args(current_report))
+  end
+
+  def recurred_for_more_than_three_days_error(current_report)
+    HistoricalLinkChecker::Error::RecurredForMoreThanThreeDaysLinkError.new(locale_interpolation_args(current_report))
+  end
+
+  def sorted_checks(checks)
+    checks.sort_by { |check| -1 * check.completed_at.to_i }
+  end
+
+  def checks_since(date)
+    checks.select { |check| check.completed_at > date }
+  end
+
+  def checks_have_errors?(checks_to_check)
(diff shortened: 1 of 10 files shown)


spec/spec_helper.rb
```diff
@@ -3,6 +3,7 @@
 require "webmock/rspec"
 require "timecop"
 require "govuk_sidekiq/testing"
+require 'rspec/its'
```
kevindew: can you maintain the double quote consistency here? (our coding standards rules regarding quotes are annoyingly loose but generally it's maintain convention of the file, on new files preference towards double quotes)

lib/historical_link_checker.rb
```diff
@@ -0,0 +1,77 @@
...
+      current_report.add_problem(
+        recurred_for_more_than_three_days_error(current_report)
+      )
+    else
+      current_report
+    end
+  end
+
+  def needs_historical_data_check?(current_report)
+    return false unless current_report.has_errors?
```
kevindew: do we care about warnings too?

lib/historical_link_checker/error/recurred_for_more_than_one_week_link_error.rb
```diff
@@ -0,0 +1,12 @@
+module HistoricalLinkChecker::Error
+  class RecurredForMoreThanOneWeekLinkError < ::LinkChecker::UriChecker::Error
```
kevindew: Can we shorten the class names at all? I don't think we have a convention of LinkError suffix on other ones.

lib/historical_link_checker.rb
```diff
@@ -0,0 +1,77 @@
...
+    else
+      current_report
+    end
+  end
+
+private
+
+  attr_reader :uri, :checks
+
+  def check_history(current_report)
```
kevindew: Not sure if missing something, but is there anything that checks whether it is the same problem that is occurring?

As far as I can tell you could have something return 500 for a week and then next return 404 and this would say 404 not found has been occurring for 7 days

lib/historical_link_checker.rb
```diff
@@ -0,0 +1,77 @@
...
+    if all_checks_have_errors?(checks_since(7.days.ago))
+      current_report.add_problem(
+        recurred_for_more_than_one_week_error(current_report)
+      )
+    elsif all_checks_have_errors?(checks_since(4.days.ago))
+      current_report.add_problem(
+        recurred_for_more_than_three_days_error(current_report)
+      )
+    else
```
kevindew: Would you end up with two errors that are for the same thing e.g.

- This problem has been occurring for more than 3 days. Page Not Found
- Page Not Found 

?

lukemorton:
I'd in agree in general that this feature is making the `link-checker-api` stateful. I guess the question is, will all consumers benefit from this behaviour? If they will, how do we make the state consistent no matter when you call the API.

It's my understanding that in addition of `local-links-manager`, this link checking behaviour is to be added to `whitehall` also. It would be useful to be able to prioritise links based on whether they are a temporal or persistent issue.

What I think the two main objection points of @kevindew are (and correct me if I'm wrong):

 - With this implementation `link-checker-api` will provide unreliable results depending on how often the caller calls to request a link checked
 - We are unsure whether monitoring should be here or in fact consumers of the API

If we are to provide the ability of prioritising links based on persistency of issues in multiple consumers, what are the alternatives to implementing this behaviour here?

I see a few options:

1. Create another link-monitoring-api that both whitehall and local-links-manager can consume
2. Build monitoring into link-checker-api
3. Build monitoring logic into whitehall and local-links-manager separately

Of course all these solutions are going to be checking thousands to millions of links on a nightly or weekly basis. We can check domains before checking URLs to reduce workload potentially.

Thoughts?

kevindew:
@lukemorton there's some discussion about this in the #link-checker-api slack channel as well - but yes you've understood my concerns correctly 👍 

With Whitehall there are two kinds of link checking to bear in mind, there is the on demand one for a publisher and then the monthly one. 

I'd agree those are the 3 options too.

I imagine you effectively have the two contexts of on-demand and monitoring in how apps are interested:

ie. If I'm in Whitehall Publisher and click "check links" there's an expectation that it goes off and checks the links there and seems somewhat confusing to have a mix of history and non-history going on.

whereas, If I'm in Local Links Manager, it could be useful to see that a link that has been 404ing since January is more of a concern than one that's been broken since yesterday.

SebAshton:
Closed in favour of #87

PR #87, opened 2017-11-08
First pass link monitoring

SebAshton:
This PR introduces:

- A new `POST` endpoint to create new monitored resources, which is a collection of links that lives within another service. e.g. Whitehall. When a monitored resource is created, it also creates a joining entity - `MonitorLink` - where we store a history of link errors. This history is buildup from the results of the existing checks.

- New workers that monitor the links associated with monitored resources. `ScheduleResourceMonitorWorker` is scheduled to run at 1 am every day [i.e. a quiet period] to queue up `ResourceMonitorWorker`'s which call `LinkMonitor::UseCase::CheckMonitoredLinks` for each enabled resource monitor.

- In order to see the result of this, we are temporarily decorating the link_report with the dates on which the error first started to appear.

### DIFF
--- a/app/controllers/monitor_controller.rb
+++ b/app/controllers/monitor_controller.rb
@@ -1,0 +1,19 @@
+class MonitorController < ApplicationController
+  def create
+    monitor = LinkMonitor::CreateResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service]).call
+
+    monitor.validate!
+
+    render(json: monitor_report(monitor), status: 200)
+  end
+
+private
+
+  def permitted_params
+    params.permit(:service, links: [])
+  end
+
+  def monitor_report(monitor)
+    { id: monitor.id }
+  end
+end
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -6,7 +6,7 @@ class Check < ApplicationRecord
 
   belongs_to :link
 
-  scope :created_within, -> (within) { where("created_at > ?", Time.now - within) }
+  scope :created_within, ->(within) { where("created_at > ?", Time.now - within) }
   scope :requires_checking, -> { where(started_at: nil).or(Check.where(completed_at: nil).where("created_at < ?", RECHECK_THRESHOLD)) }
 
   def self.fetch_all(links, within: 4.hours)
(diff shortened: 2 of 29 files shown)


app/controllers/monitor_controller.rb
```diff
@@ -0,0 +1,21 @@
+class MonitorController < ApplicationController
+  def create
+    monitor = LinkMonitor::UseCase::CreateResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service]).call
+
+    if monitor.valid?
+      render(json: monitor_report(monitor), status: 200)
+    else
+      render(json: monitor.errors, status: 422)
```
thomasleese: In other controllers we've just called `monitor.validate!` and left Rails to handle the output for us. Perhaps we can do the same here?

app/models/check.rb
```diff
@@ -53,4 +53,28 @@ def status
...
+    if link.monitor_link.present?
+      combine_link_history
+    else
+      link_errors
+    end
+  end
+
+private
+
+  def combine_link_history
```
thomasleese: Might be nice to make this generic enough to work on the errors and the warnings.

app/presenters/monitor_presenter.rb
```diff
@@ -0,0 +1,7 @@
+class MonitorPresenter < SimpleDelegator
```
thomasleese: Does this get used anywhere?

lib/link_monitor/use_case/check_monitored_links.rb
```diff
@@ -0,0 +1,50 @@
+module LinkMonitor
+  module UseCase
```
thomasleese: I'm not quite sure why we have a `UseCase` module here?

app/models/check.rb
SebAshton: For now we're only worrying about errors, integration with warnings is on our backlog though

app/presenters/monitor_presenter.rb
SebAshton: Just tidied this up.

lib/link_monitor/use_case/check_monitored_links.rb
SebAshton: We've unwrapped the classes from this module

app/controllers/monitor_controller.rb
SebAshton: 👍

app/models/check.rb
```diff
@@ -53,4 +53,28 @@ def status
...
+    link_errors.reduce([]) do |array, error|
+      error_history = link.monitor_link.link_errors.select { |link_error| link_error['message'] == error }
+
+      array << if error_history
```
lukemorton: separate method for this block?

lib/link_monitor/create_resource_monitor.rb
```diff
@@ -0,0 +1,29 @@
...
+  class CreateResourceMonitor
+    def initialize(links:, service:)
+      @links = links
+      @service = service
+    end
+
+    def call
+      monitor = ResourceMonitor.create(service: service)
+
+      if monitor.valid?
```
lukemorton: one liner?

PR #88, opened 2017-11-10
Fix exception in error history lookup

SebAshton:


lukemorton:
Couple of things here:

- There are two separate bugs being fixed here, probably two PRs?
- There are no tests failing because of these bugs, we should add failings tests and make them green by fixing the bugs

spec/presenters/check_presenter_spec.rb
```diff
@@ -20,6 +20,15 @@
 
   it { is_expected.to eq(expected_link_report) }
 
+  context 'when the link has a monitored link but no historical checks yet' do
```
thomasleese: Would be good to stick with `"` which is our style across all our apps.

PR #89, opened 2017-11-14
Add Resource Monitor Upserting

SebAshton:


lib/link_monitor/upsert_resource_monitor.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+  private
+
+    attr_reader :links, :service, :resource_type, :resource_id
+
+    def create_links(monitor)
+      links.each do |link|
+        monitored_link = Link.find_or_create_by(uri: link)
+        monitor.monitor_links.find_or_create_by(link: monitored_link)
```
lukemorton: Is there any eventuality that would require a link to be updated during this call? We currently handle inserting and deleting, but not updating.

lib/link_monitor/upsert_resource_monitor.rb
```diff
@@ -0,0 +1,42 @@
...
+  class UpsertResourceMonitor
+    def initialize(links:, service:, resource_type:, resource_id:)
+      @links = links
+      @service = service
+      @resource_type = resource_type
+      @resource_id = resource_id
+    end
+
+    def call
+      monitor = ResourceMonitor.find_or_create_by(service: service, resource_type: resource_type, resource_id: resource_id)
```
lukemorton: We do a find by 3 different properties here, do we expect the service or resource type to change but still want to use the old ResourceMonitor record?

With this current implementation, will there be instances where we create a new ResourceMonitor and do not clean up and old one?

spec/acceptance/upserting_resource_monitor_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  context "when we create another resource monitor for an existing resource monitor" do
+    let(:resource) { subject }
+
+    before do
+      resource
+
+      links << "http://example.com/c"
+      links.shift
```
lukemorton: This is removing `/a` from the links array but I reckon we could make this easier to comprehend.

This might be a little clearer:

``` ruby
links.delete("http://example.com/a")
```

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
```diff
@@ -0,0 +1,6 @@
+class AddResourceTypeAndResourceIdToResourceMonitor < ActiveRecord::Migration[5.0]
+  def change
+    add_column :resource_monitors, :resource_type, :string
+    add_column :resource_monitors, :resource_id, :integer, index: true
```
lukemorton: We've prefixed these fields with `resource_` even though we have the word resource in the model and table name. Seems a little verbose.

app/controllers/monitor_controller.rb
```diff
@@ -1,6 +1,6 @@
 class MonitorController < ApplicationController
   def create
-    monitor = LinkMonitor::CreateResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service]).call
+    monitor = LinkMonitor::UpsertResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service], resource_type: permitted_params[:resource_type], resource_id: permitted_params[:resource_id]).call
```
thomasleese: Might be worth making `permitted_params` return using `.to_h` and then can just do `UpsertResourceMonitor.new(**permitted_params)` here.

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
thomasleese: I agree, it seems like we should rename `service` to `application`, and then we can have `resource_type` as `name` and then we don't need the `resource_id` field.

thomasleese: Probably worth making these `null: false` as well.

spec/acceptance/checking_monitored_resources_spec.rb
```diff
@@ -24,9 +24,11 @@
 
   context 'when another service monitors the same link' do
     let(:other_resource_monitor) do
-      LinkMonitor::CreateResourceMonitor.new(
+      LinkMonitor::UpsertResourceMonitor.new(
         links: resource_monitor.links.map(&:uri),
-        service: "local-link-manager"
+        service: "local-link-manager",
```
thomasleese: Incredibly small thing! But, might be confusing when doing a search through the code, so this should be `local-links-manager`.

lib/link_monitor/upsert_resource_monitor.rb
thomasleese: It seems to me that applications shouldn't affect the `ResourceMonitor`s of others, so `service` makes sense to me here. But I don't quite see the advantage of having both a `resource_type` and a `resource_id`. It feels like we can have a `service` and a `name`.

lukemorton: ResourceMonitor will be per edition on whitehall, so need to uniquely identify them which is why resource_id probably exists.

If `application` is used over `service` in the GDS world we should probably be consistent with that.

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
lukemorton: If `application` is used over `service` in the GDS world we should probably be consistent with that.

lib/link_monitor/upsert_resource_monitor.rb
thomasleese: Hmm, this feels to me like Link Checker API knowing too much about the caller of the API. I imagine this could work by just having Whitehall provide a `name` (or `resource_type`) of `edition:4322` - and similar for Local Links Manager.

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
thomasleese: We have `publishing_app` and `rendering_app` in the Publishing APP, so `app` might be the better choice. Although in Router we have `backend_id`...

lib/link_monitor/upsert_resource_monitor.rb
lukemorton: Yeah I'd tend to agree, maybe `reference` would be better than `name`, but yeah, leave it up to the caller to ensure it's unique.

thomasleese: Yeah `reference` works well, it's what GOV.UK Notify uses in fact.

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
SebAshton: Ok, so we're going with `app` and `reference` ?

spec/acceptance/checking_monitored_resources_spec.rb
SebAshton: 👍

PR #90, opened 2017-11-14
Update recourse monitor upsert endpoint nomenclature

SebAshton:
Takes onboard comments from #89

### DIFF
--- a/app/controllers/monitor_controller.rb
+++ b/app/controllers/monitor_controller.rb
@@ -1,6 +1,9 @@
 class MonitorController < ApplicationController
   def create
-    monitor = LinkMonitor::UpsertResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service], resource_type: permitted_params[:resource_type], resource_id: permitted_params[:resource_id]).call
+    monitor = LinkMonitor::UpsertResourceMonitor.new(
+      links: permitted_params[:links],
+      app: permitted_params[:app], reference: permitted_params[:reference]
+    ).call
 
     monitor.validate!
 
@@ -10,7 +13,7 @@ class MonitorController < ApplicationController
 private
 
   def permitted_params
-    params.permit(:service, :resource_type, :resource_id, links: [])
+    params.permit(:app, :reference, links: [])
   end
 
   def monitor_report(monitor)
--- a/app/models/resource_monitor.rb
+++ b/app/models/resource_monitor.rb
@@ -1,7 +1,6 @@
 class ResourceMonitor < ApplicationRecord
-  validates :service, uniqueness: { scope: %i(resource_type resource_id) },
-                      presence: true
-  validates_presence_of :resource_type, :resource_id
+  validates :app, presence: true
+  validates :reference, uniqueness: { scope: :app }, presence: true
 
   has_many :monitor_links
   has_many :links, through: :monitor_links
--- a/db/migrate/20171115105030_add_reference_to_resource_monitor.rb
+++ b/db/migrate/20171115105030_add_reference_to_resource_monitor.rb
@@ -1,0 +1,6 @@
+class AddReferenceToResourceMonitor < ActiveRecord::Migration[5.0]
+  def change
+    add_column :resource_monitors, :reference, :string, default: 'bar', null: false, index: true
+    change_column_default :resource_monitors, :reference, nil
+  end
+end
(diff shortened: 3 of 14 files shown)


app/controllers/monitor_controller.rb
```diff
@@ -1,6 +1,6 @@
 class MonitorController < ApplicationController
   def create
-    monitor = LinkMonitor::UpsertResourceMonitor.new(links: permitted_params[:links], service: permitted_params[:service], resource_type: permitted_params[:resource_type], resource_id: permitted_params[:resource_id]).call
+    monitor = LinkMonitor::UpsertResourceMonitor.new(links: permitted_params[:links], app: permitted_params[:app], reference: permitted_params[:reference]).call
```
thomasleese: Would be good to make this on multiple lines to improve readability.

app/models/resource_monitor.rb
```diff
@@ -1,7 +1,6 @@
 class ResourceMonitor < ApplicationRecord
-  validates :service, uniqueness: { scope: %i(resource_type resource_id) },
-                      presence: true
-  validates_presence_of :resource_type, :resource_id
+  validates :app, uniqueness: { scope: :reference }, presence: true
```
thomasleese: Don't think it actually matters, but this might be better expressed as "the reference must be unique among the apps" rather than currently where it's saying "the app must be unique among the references".

db/migrate/20171113145621_add_resource_type_and_resource_id_to_resource_monitor.rb
```diff
@@ -1,6 +1,5 @@
 class AddResourceTypeAndResourceIdToResourceMonitor < ActiveRecord::Migration[5.0]
   def change
-    add_column :resource_monitors, :resource_type, :string
-    add_column :resource_monitors, :resource_id, :integer, index: true
+    add_column :resource_monitors, :reference, :string, null: false, index: true
```
thomasleese: I think you'll need to put this into a separate migration since this one will have already run on integration.

db/schema.rb
```diff
@@ -60,12 +60,11 @@
...
 
   create_table "resource_monitors", force: :cascade do |t|
-    t.boolean  "enabled",       default: true
-    t.string   "service"
-    t.datetime "created_at",                   null: false
-    t.datetime "updated_at",                   null: false
-    t.string   "resource_type"
-    t.integer  "resource_id"
+    t.boolean  "enabled",    default: true
+    t.string   "app"
```
thomasleese: Would be nice to get this `null: false` as well.

SebAshton:
@thomasleese should be GTM

db/migrate/20171115105030_add_reference_to_resource_monitor.rb
```diff
@@ -0,0 +1,5 @@
+class AddReferenceToResourceMonitor < ActiveRecord::Migration[5.0]
+  def change
+    add_column :resource_monitors, :reference, :string, null: false, index: true
```
thomasleese: Do we know if there are any resource monitors in the database already which will cause a `null` constraint error when the column is created? (unfortunately integration is looking a bit unhealthy right now so I can't check)

SebAshton: We created a couple of test monitors last week. So possibly if it has not been synced with staging/prod since then. Could throw in a `ResourceMonitor.all.destroy` ?

thomasleese: I would add the column as nullable, fill it with nonsense references and then make the column not nullable - and same for `app` I guess.

db/schema.rb
```diff
@@ -60,12 +60,11 @@
   end
 
   create_table "resource_monitors", force: :cascade do |t|
-    t.boolean  "enabled",       default: true
-    t.string   "service"
-    t.datetime "created_at",                   null: false
-    t.datetime "updated_at",                   null: false
-    t.string   "resource_type"
-    t.integer  "resource_id"
+    t.boolean  "enabled",    default: true
```
thomasleese: Sorry, one more thing! Could we get this not nullable as well?

SebAshton: What benefit would the system have by making this nullable?

thomasleese: Without `null: false`, `boolean` fields can be true, false or null. It would be better that they can only be true or false.

SebAshton: OIC 👍  made the change

PR #91, opened 2017-11-15
Check existing links are not updated

tsteer:


rubocop-d89603d.html
```diff
@@ -0,0 +1,385 @@
+<!DOCTYPE html>
```
thomasleese: I think this accidentally got added in this commit! :)

spec/acceptance/upserting_resource_monitor_spec.rb
```diff
@@ -70,6 +71,10 @@
     it "should not destroy links" do
       expect(Link.all.count).to eq(3)
     end
+
+    it 'should not update any existing links' do
```
thomasleese: We generally like to stick with double quotes on strings in GDS.

PR #94, opened 2017-11-16
Add organisation tags to resource monitor

tsteer:
Organisation tags have been added so we can find the history of link errors by an organisation


thomasleese:
Would it be possible to add a description to the commit here to explain the need for this?

thomasleese:
Thanks for adding a description. I'm still not quite sure of the benefit of including this into the Link Checker API directly; to me, this feels like something which could belong in the clients.

Perhaps an alternative might be to allow clients to store arbitrary user data along with a resource monitor which leaves it up to the client to decide how best to use it.

tsteer:
Based on the discussions that were had yesterday. I'm going to close this now, and reopen it when we need it - when we create the new service for viewing broken links

PR #98, opened 2017-11-21
Add link history

tsteer:
Added link history table - to store error histories for each link. Migration to add the link_histories table, migration to copy data from monitor links to link_histories and migration to delete link errors from monitor_links

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,6 +1,7 @@
 class Link < ApplicationRecord
   has_many :checks
   has_many :monitor_links
+  has_one :link_history
 
   validates_presence_of :uri
 
--- a/app/models/link_history.rb
+++ b/app/models/link_history.rb
@@ -1,0 +1,34 @@
+class LinkHistory < ApplicationRecord
+  belongs_to :link
+
+  #To do: store link warnings
+  def update_errors(errors)
+    remove_resolved_errors(errors)
+    errors.each { |error| add_error(error) }
+  end
+
+  def add_error(message)
+    return if error_exists?(message)
+
+    link_errors << {
+      message: message,
+      started_at: DateTime.now
+    }
+
+    save
+  end
+
+  def clear_errors
+    update!(link_errors: [])
+  end
+
+private
+
+  def remove_resolved_errors(errors)
+    link_errors.keep_if { |link_error| errors.include?(link_error[:message]) }
+  end
+
+  def error_exists?(message)
+    link_errors.any? { |error| error[:message] == message }
+  end
+end
--- a/db/migrate/20171121110841_create_link_histories.rb
+++ b/db/migrate/20171121110841_create_link_histories.rb
@@ -1,0 +1,10 @@
+class CreateLinkHistories < ActiveRecord::Migration[5.0]
+  def change
+    create_table :link_histories do |t|
+      t.json :link_errors, default: [], null: false
+      t.references :link, index: true, foreign_key: true, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20171121113257_migrate_data_from_monitor_link_to_link_history.rb
+++ b/db/migrate/20171121113257_migrate_data_from_monitor_link_to_link_history.rb
@@ -1,0 +1,10 @@
+class MigrateDataFromMonitorLinkToLinkHistory < ActiveRecord::Migration[5.0]
+  def change
+    MonitorLink.find_each do |link|
+      LinkHistory.create(
+        link_id: link.id,
+        link_errors: link.link_errors
+      )
+    end
+  end
+end
--- a/db/migrate/20171123105000_remove_last_checked_at_from_monitor_links.rb
+++ b/db/migrate/20171123105000_remove_last_checked_at_from_monitor_links.rb
@@ -1,0 +1,5 @@
+class RemoveLastCheckedAtFromMonitorLinks < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :monitor_links, :last_checked_at, :datetime
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171121110841) do
+ActiveRecord::Schema.define(version: 20171121113257) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171121120032) do
+ActiveRecord::Schema.define(version: 20171123105000) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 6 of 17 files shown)


db/migrate/20171121110841_create_link_histories.rb
```diff
@@ -0,0 +1,10 @@
+class CreateLinkHistories < ActiveRecord::Migration[5.0]
+  def change
+    create_table :link_histories do |t|
+      t.json :link_errors, default: []
+      t.belongs_to :link
```
thomasleese: Can we use `t.references :link, index: true, foreign_key: true` to get the foreign key as well?

db/migrate/20171121113257_migrate_data_from_monitor_link_to_link_history.rb
```diff
@@ -0,0 +1,10 @@
+class MigrateDataFromMonitorLinkToLinkHistory < ActiveRecord::Migration[5.0]
+  def change
+    MonitorLink.all.each do |link|
```
thomasleese: Can we use https://apidock.com/rails/ActiveRecord/Batches/find_each to make this more efficient?

db/schema.rb
```diff
@@ -44,6 +44,14 @@
     t.index ["link_id"], name: "index_checks_on_link_id", using: :btree
   end
 
+  create_table "link_histories", force: :cascade do |t|
+    t.json     "link_errors", default: []
+    t.integer  "link_id"
```
thomasleese: Would be nice to make this `null: false`.

db/schema.rb
```diff
@@ -44,6 +44,14 @@
     t.index ["link_id"], name: "index_checks_on_link_id", using: :btree
   end
 
+  create_table "link_histories", force: :cascade do |t|
+    t.json     "link_errors", default: []
```
thomasleese: Would be nice to make this `null: false`.

app/models/link.rb
```diff
@@ -1,6 +1,7 @@
 class Link < ApplicationRecord
   has_many :checks
   has_many :monitor_links
```
thomasleese: Does this still need to be a `has_many` following this change?

app/models/link_history.rb
```diff
@@ -0,0 +1,24 @@
+class LinkHistory < ApplicationRecord
+  belongs_to :link
+
+  def add_errors(errors)
```
thomasleese: Do we need to think about the warnings as well as the errors?

app/models/link_history.rb
```diff
@@ -0,0 +1,24 @@
...
+    link_errors.keep_if { |link_error| errors.include?(link_error[:message]) }
+
+    errors.each { |error| add_error(error) }
+  end
+
+  def add_error(message)
+    return if link_errors.any? { |error| error[:message] == message }
```
thomasleese: Might be good to extract this into a `has_error?`.

app/models/link_history.rb
```diff
@@ -0,0 +1,24 @@
+class LinkHistory < ApplicationRecord
+  belongs_to :link
+
+  def add_errors(errors)
+    link_errors.keep_if { |link_error| errors.include?(link_error[:message]) }
```
thomasleese: We can avoid doing this check and just leave `add_error` to do this.

spec/acceptance/checking_monitored_resources_spec.rb
```diff
@@ -3,7 +3,11 @@
 RSpec.describe "Check all links for a monitored resource" do
   let(:resource_monitor) { FactoryGirl.create(:resource_monitor, number_of_links: 1) }
 
-  let(:report) { LinkChecker::UriChecker::Report.new }
+  let(:report) do
+    LinkChecker::UriChecker::Report.new.add_problem(
+      TestError::PageNotFound.new(from_redirect: false)
```
thomasleese: Might be good to move the `add_problem` part into a `before` block.

app/models/link.rb
tsteer: Yep, its a many to many relationship :)

app/models/link_history.rb
tsteer: At the moment we are just focusing on the errors, but its in our backlog, so will look into it then

db/schema.rb
```diff
@@ -54,7 +62,7 @@
     t.integer  "link_id"
     t.integer  "resource_monitor_id"
     t.datetime "last_checked_at"
-    t.json     "link_errors",         default: []
+    t.json     "link_errors"
```
thomasleese: I think this is an erroneous change.

app/models/link_history.rb
thomasleese: Hmm, okay. I wonder if it's worth putting a comment in to explain this as I'm a bit worried that feature will be forgotten.

spec/models/link_history_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe LinkHistory, type: :model do
+  describe "#add_errors" do
+    let(:link) { FactoryGirl.create(:link) }
+    let(:link_history) { FactoryGirl.create(:link_history, link: link) }
```
thomasleese: We could move the `link: link` part here into the factory so we don't need to explicitly create a link.

app/models/link_history.rb
tsteer: This line was removing any resolved errors from link_errors. We extracted it into a separate method for clarity.

thomasleese: Oh I see, perhaps we should rename the method to `update_errors` to make it clearer that it doesn't just add errors.

app/models/link_history.rb
```diff
@@ -0,0 +1,41 @@
...
+    save
+  end
+
+  def clear_errors
+    update!(link_errors: [])
+  end
+
+private
+
+  #remove_resolved_errors removes any resolved errors from link_errors
```
thomasleese: Ah oops, I didn't mean to imply that we needed comments here, I meant we should put a comment about why warnings haven't been included in this class yet and rename `add_errors` to `update_errors` to make it clear it does more than just adding.

lib/link_monitor/check_monitored_links.rb
```diff
@@ -17,19 +17,17 @@ def check_link(link)
...
+      update_link_history(link)
     end
 
-    def update_monitor_link(link)
-      monitor_link = resource_monitor.monitor_links.find_by(link_id: link.id)
+    def update_link_history(link)
+      link_history = link.link_history ||= link.create_link_history
       check = link.checks.order(completed_at: :desc).first
 
-      monitor_link.touch(:last_checked_at)
```
thomasleese: If we're no longer updating this value, can we get rid of the column?

db/migrate/20171123105000_remove_last_checked_at_from_monitor_links.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveLastCheckedAtFromMonitorLinks < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :monitor_links, :last_checked_at, :datetime
```
thomasleese: You'll also need to remove the field from https://github.com/alphagov/link-checker-api/blob/master/spec/factories/monitor_links.rb#L3 - might be worth adding a test so this would have been caught.

tsteer: What do you think the value of the test would be? Just because it seems like it wouldn't be testing the behaviour, rather the absence of a field?

thomasleese: I was just surprised because it means there is no test which creates a `MonitorLink` factory, so if another developer tries to use that factory it will fail and could be confusing as to why.

A pattern we've used in a number of apps is to create a test that just ensures the default factory is valid: https://github.com/alphagov/email-alert-api/blob/master/spec/models/subscriber_spec.rb#L7

PR #100, opened 2017-12-05
Upgrade to Ruby 2.4.2 & Bundle update

danielburnley:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.

https://trello.com/c/vX8qwh2X/46-ruby-upgrade-of-all-apps


lib/link_checker/uri_checker/valid_uri_checker.rb
```diff
@@ -11,7 +11,7 @@ def initialize(options = {})
     end
   end
 
-  class ContactDetails < Warning
+  class ContactDetails < LinkChecker::UriChecker::Warning
```
thomasleese: Why is this needed on only one of the classes and not the others?

danielburnley: Without it you get: 

```
Failure/Error:                                                                                      
        class ContactDetails < Warning                                                                    
          def initialize(options = {})                                                                    
            super(summary: :contact_details, message: :links_to_contact_details, suggested_fix: :check_correct_manually, **options)                                                                                  
          end                                                                                             
                                                                                                          
      TypeError:                                                                                          
        superclass must be a Class (Module given)
```

I'm still trying to find out as to _why_ this occurs though. Any thoughts @thomasleese ?

danielburnley: Ruby 2.4 introduced the `Warning` module - which is probably what it thinks it's being inherited from there. As to why it's not happening in the other places this is inherited I'm not sure, might be a weird loading issue. I'll inherit from the correct place explicitly in every occurrence though just in case

thomasleese: That's annoying, but seems reasonable. Good to put it explicitly in all places.

kevindew: If you did a require at the top you'd probably not need to do anything there - wonder if it's a rails issue?

Similarly could loosen it to class `ContactDetails < UriChecker::Warning` - probably also good to make Error consistent with it.

PR #102, opened 2017-12-12
Enable cookies to avoid infinite redirect loops

tsteer:
This change adds cookies to Faraday to avoid false positives being raised by HttpChecker. False positives were occuring because some web pages redirect to themselves if no cookies are present. Because we never store cookies whilst checking, the check raises it as a redirect loop error incorrectly.

Reproduce the issue with cURL:

```
curl -I -L 'https://sharedsystems.eu-supply.com/login.asp?B=SELLAFIELD'
```

Fix the issue with cURL:

```
curl --cookies -I -L 'https://sharedsystems.eu-supply.com/login.asp?B=SELLAFIELD'
```

We move the initialisation of the Faraday connection object into a class level singleton so the cookie jar can be shared. Cookie jar sharing needs to occur across instances of the class because a new instance of HttpChecker is created for each redirect.

## Lack of tests

We can create a test that replicates this issue if we disable Webmock. Unfortunately the issue cannot be recreated with Webmock and therefore we haven't got a test with this change.

Any thoughts on how we can test it?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem "pg"
 gem "govuk_app_config", "~> 0.2"
 
 gem "faraday", "~> 0.11"
+gem "faraday-cookie_jar", "~>0.0.6"
 
 gem "govuk_sidekiq", "~> 2.0"
 gem "sidekiq-scheduler", "~> 2.1"
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -92,6 +92,7 @@ module LinkChecker::UriChecker
   class HttpChecker < Checker
     def self.connection
       @connection ||= Faraday.new(headers: { accept_encoding: "none" }) do |faraday|
+        faraday.use :cookie_jar
         faraday.adapter Faraday.default_adapter
       end
     end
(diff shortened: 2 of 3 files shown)


lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -119,13 +128,14 @@ def call
...
     REDIRECT_LIMIT = 8
+    REDIRECT_LOOP_LIMIT = 1
     REDIRECT_WARNING = 2
     RESPONSE_TIME_LIMIT = 15
     RESPONSE_TIME_WARNING = 5
 
     def check_redirects
       add_problem(TooManyRedirects.new(from_redirect: from_redirect?)) if redirect_history.length >= REDIRECT_LIMIT
-      add_problem(RedirectLoop.new(from_redirect: from_redirect?)) if redirect_history.include?(uri)
+      add_problem(RedirectLoop.new(from_redirect: from_redirect?)) if redirect_history.count(uri) > REDIRECT_LOOP_LIMIT
```
kevindew: I think it would be better if this constant was set to 2 and followed the same `>=` behaviour of the line above, otherwise the limit values mean something slightly different.

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -1,3 +1,5 @@
+require "faraday-cookie_jar"
```
kevindew: I'd have thought this gets already required as part of the gem - just wondering if it's needed

Gemfile
```diff
@@ -12,6 +12,7 @@ gem "pg"
 gem "govuk_app_config", "~> 0.2"
 
 gem "faraday", "~> 0.11"
+gem 'faraday-cookie_jar'
```
thomasleese: Would be good to get this with a version like "~> 0.0"`.

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -90,6 +92,13 @@ def initialize(options = {})
   end
 
   class HttpChecker < Checker
+    def self.connection
```
thomasleese: Would be good to get this refactor in a separate commit.

tsteer:
@kevindew @thomasleese Updated :)

PR #103, opened 2017-12-13
Fix redirect loop caused by lack of redirect leeway

tsteer:
Some websites require multiple refreshes to the same page to generate cookies required for page viewing. This fix increases the number of page redirects allowed to ensure that all cookies can be generated to view the page. Redirect limit set to 5.


tsteer:
@kevindew @thomasleese Could you review this please

PR #113, opened 2018-01-05
Link report rake task

thomasleese:
This adds a new rake task which generates a report of all broken links and when they were last checked. This was requested to be able to cross reference this with the data in Google Analytics to identify how many users are affected by broken links.

### DIFF
--- a/app/services/report_generator_service.rb
+++ b/app/services/report_generator_service.rb
@@ -1,0 +1,37 @@
+require "csv"
+
+class ReportGeneratorService
+  def initialize(filename)
+    @filename = filename
+  end
+
+  def call
+    CSV.open(filename, "w") do |csv|
+      csv << csv_headings
+      checks_with_errors.each do |check|
+        csv << csv_row_for_check(check)
+      end
+    end
+  end
+
+private
+
+  attr_reader :filename
+
+  def checks_with_errors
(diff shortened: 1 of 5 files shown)


app/services/report_generator_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+class ReportGeneratorService
+  def initialize(filename)
+    @filename = filename
+  end
+
+  def call
+    CSV.open(filename, "w") do |csv|
+      csv << csv_headings
+      checks_with_errors.map do |check|
```
kevindew: Not a biggie but did you mean an each here?

app/services/report_generator_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  attr_reader :filename
+
+  def checks_with_errors
+    Link.includes(:checks)
+      .map(&:most_recent_check)
+      .compact
+      .select do |check|
+        check.completed? && check.has_errors?
+      end
```
kevindew: Not sure how concerned you are with performance here, but assuming there's a lot of checks per link (I can't remember quite how it works) you might want to switch this round to a query on check rather than links.

kevindew: okey doke 👍

PR #117, opened 2018-01-09
Lock created Batch within transaction

SebAshton:
This PR locks the created batch within the Active record transaction.


thomasleese:
Should be solved by https://github.com/alphagov/whitehall/pull/3675.

PR #118, opened 2018-01-09
[WIP] Extract transaction into separate method

SebAshton:
This PR is one of two potential solutions for a concurrency issue we are seeing when creating a lot of batches. The other being #117.

### DIFF
--- a/app/controllers/batch_controller.rb
+++ b/app/controllers/batch_controller.rb
@@ -26,22 +26,9 @@ class BatchController < ApplicationController
     create_params = CreateParams.new(params)
     create_params.validate!
 
-    batch = ActiveRecord::Base.transaction do
-      links = Link.fetch_all(create_params.uris)
-      checks = Check.fetch_all(links, within: create_params.checked_within)
-      batch = Batch.create!(
-        webhook_uri: create_params.webhook_uri,
-        webhook_secret_token: create_params.webhook_secret_token
-      )
+    batch = create_batch(create_params)
 
-      batch_checks = checks.each_with_index.map do |check, i|
-        BatchCheck.new(batch_id: batch.id, check_id: check.id, order: i)
-      end
-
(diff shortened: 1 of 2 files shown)


app/controllers/batch_controller.rb
```diff
@@ -62,6 +49,29 @@ def show
 
 private
 
+  def create_batch(create_params)
+    batch = Batch.create!(
```
thomasleese: Can we not still create the batch inside the transaction?

app/controllers/batch_controller.rb
```diff
@@ -26,22 +26,9 @@ def create
...
 
-      batch_checks = checks.each_with_index.map do |check, i|
-        BatchCheck.new(batch_id: batch.id, check_id: check.id, order: i)
-      end
-
-      BatchCheck.import(batch_checks)
-
-      batch
-    end
+    head :unprocessable_entity && return unless batch
```
thomasleese: This sounds like a good idea, but I would perhaps write it as:

```ruby
return head :unprocessable_entity unless batch
```

app/controllers/batch_controller.rb
```diff
@@ -62,6 +49,29 @@ def show
...
+
+      batch_checks = checks.each_with_index.map do |check, i|
+        BatchCheck.new(batch_id: batch.id, check_id: check.id, order: i)
+      end
+
+      BatchCheck.import(batch_checks)
+    end
+
+    batch
+  rescue ActiveRecord::StatementInvalid, ActiveRecord::Rollback, ActiveRecord::RecordInvalid
```
thomasleese: I assume any of these exceptions would cause the transaction to be automatically rolled back anyway.

thomasleese:
Should be solved by https://github.com/alphagov/whitehall/pull/3675.

PR #119, opened 2018-01-11
Re enable google safebrowsing

SebAshton:
Now we have a Google SafeBrowsing API key in gov-secrets we can now re-enable the SafeBrowsing API check.

### DIFF
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -105,9 +105,7 @@ module LinkChecker::UriChecker
       check_meta_mature_rating
       return report if report.has_errors?
 
-      # disabled for the moment until we can sort out a process for getting
-      # permentant staging and production keys
-      # check_google_safebrowsing
+      check_google_safebrowsing if use_google_safebrowsing?
 
       report
     end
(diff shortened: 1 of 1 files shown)


lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -180,7 +178,7 @@ def check_meta_mature_rating
 
     def check_google_safebrowsing
       api_key = Rails.application.secrets.google_api_key
-      return unless api_key
+      return if !Rails.env.production? || !api_key
```
thomasleese: I think we should move this check above to where we call it and then raise an exception here if the `api_key` is not set in the method.

```ruby
check_google_safebrowsing unless Rails.env.production? || Rails.application.secrets.google_api_key
```

```ruby
raise "Missing API key" unless api_key
```

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -290,5 +289,9 @@ def additional_connection_headers
 
       { 'Rate-Limit-Token': Rails.application.secrets.rate_limit_token }
     end
+
+    def use_google_safebrowsing?
+      Rails.env.production? && Rails.application.secrets.google_api_key
```
thomasleese: This should be an `||`.

SebAshton: I was going down the production only route, but makes sense if you want to test locally.

SebAshton: Done @thomasleese

PR #120, opened 2018-01-15
Enables bypassing of the authentication on integration

tsteer:
This PR adds the functionality to bypass the HTTP Auth challenge on integration.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,6 +17,9 @@ require "rails/test_unit/railtie"
 Bundler.require(*Rails.groups)
 
 module LinkCheckerApi
+  mattr_accessor :hosts_with_basic_authorization
+  self.hosts_with_basic_authorization = {}
+
   class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -14,12 +14,13 @@ development:
   secret_key_base: 36746f0eb3c57b16be0278bdf9a9418ec225ecd5f9a3183697a3dc0dfd3aa1d05d285b3ed42b77c92ae6deee7af6a2b9bc87f119eca9c74810dcc43ddd5eb02b
   google_api_key: <%= ENV["GOOGLE_API_KEY"] %>
   rate_limit_token: pfB6uNKYC8sB9PVgBLdwFToN
+  govuk_basic_auth_credentials: "test:test"
 
 test:
   secret_key_base: b81a573d7703bfdf5c80f2a201c9e9d7ec5aad3326e252534568e1a82a7fc79a09f0d8ebd5e7e0fb5fc584f6a23440f2be5c986ec723df389ef7527e3e7ffc12
   google_api_key: test
   rate_limit_token: pfB6uNKYC8sB9PVgBLdwFToN
-
+  govuk_basic_auth_credentials: "test:test"
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
@@ -27,3 +28,4 @@ production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   google_api_key: <%= ENV["GOOGLE_API_KEY"] %>
   rate_limit_token: <%= ENV["RATE_LIMIT_TOKEN"] %>
+  govuk_basic_auth_credentials: <%= ENV["GOVUK_BASIC_AUTH_CREDENTIALS"] %>
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -288,16 +288,30 @@ module LinkChecker::UriChecker
       uri.path.starts_with? "/government/uploads"
     end
 
-    def additional_connection_headers
-      return nil unless gov_uk_uri?
+    def rate_limit_header
+      return {} unless gov_uk_uri?
+      { "Rate-Limit-Token": Rails.application.secrets.rate_limit_token }
+    end
 
-      { 'Rate-Limit-Token': Rails.application.secrets.rate_limit_token }
+    def basic_authorization_header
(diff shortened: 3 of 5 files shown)


config/initializers/authed_domains.rb
```diff
@@ -0,0 +1,5 @@
+if Rails.application.secrets.basic_auth_credentials
+  LinkCheckerApi.authed_domains = {
```
thomasleese: Perhaps this should just set the key in the hash rather than replacing the entire hash.

config/secrets.yml
```diff
@@ -14,16 +14,18 @@ development:
...
-
+  basic_auth_credentials: c81a772d7703bfdf8cfc79a0
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   google_api_key: <%= ENV["GOOGLE_API_KEY"] %>
   rate_limit_token: <%= ENV["RATE_LIMIT_TOKEN"] %>
+  basic_auth_credentials: <%= ENV["BASIC_AUTH_CREDENTIALS"] %>
```
thomasleese: I would be tempted to call this `govuk_basic_auth_credentials`

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -284,10 +284,22 @@ def gov_uk_uri?
...
     def additional_connection_headers
-      return nil unless gov_uk_uri?
+      headers = {}
+
+      if gov_uk_uri?
+        headers["Rate-Limit-Token"] = Rails.application.secrets.rate_limit_token
+      end
+
+      if authentication_required?
+        headers["Authorization"] = "Basic #{Rails.application.secrets.basic_auth_credentials}"
```
thomasleese: I imagine we would want to use the value in `authed_domains` here.

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -284,10 +284,22 @@ def gov_uk_uri?
       Plek.new.website_root.include?(uri.host)
     end
 
+    def authentication_required?
+      LinkCheckerApi.authed_domains[uri.host]
```
thomasleese: Might be clearer to do `basic_authorization_required?`.

lib/link_checker/uri_checker/http_checker.rb
thomasleese: I wonder if we should do something along the lines of:

```ruby
def basic_authorization_header
  return {} unless LinkCheckerApi.authed_domains.include?(uri.host)
  {"Authorization": LinkCheckerApi.authed_domains[uri.host]}
end

def additional_connection_headers
  ...
  headers.merge(basic_authorization_header
end
```

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -284,10 +284,22 @@ def gov_uk_uri?
...
+    end
+
+    def basic_authorization_header
+      return {} unless LinkCheckerApi.authed_domains.include?(uri.host)
+      { "Authorization": LinkCheckerApi.authed_domains[uri.host] }
+    end
+
     def additional_connection_headers
-      return nil unless gov_uk_uri?
+      headers = {}
```
thomasleese: We could possibly make this more succinct with:

```ruby
def ...
  Hash.new
    .merge(rate_limit_header)
    .merge(basic_authorization_header)
end
```

SebAshton: YES! @lukemorton and I were looking at this and thinking "how can we make this prettier"

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -284,10 +284,20 @@ def gov_uk_uri?
...
-      return nil unless gov_uk_uri?
+    def rate_limit_header
+      return {} unless gov_uk_uri?
+      { "Rate-Limit-Token": Rails.application.secrets.rate_limit_token }
+    end
 
-      { 'Rate-Limit-Token': Rails.application.secrets.rate_limit_token }
+    def basic_authorization_header
+      return {} unless LinkCheckerApi.authed_domains.include?(uri.host)
+      { "Authorization": LinkCheckerApi.authed_domains[uri.host] }
```
thomasleese: I think we should rename `authed_domains` to `basic_authorization_headers` and then in this method insert the `Basic ` at the start of the string.

SebAshton:
@thomasleese have made those changes

config/secrets.yml
```diff
@@ -14,16 +14,18 @@ development:
   secret_key_base: 36746f0eb3c57b16be0278bdf9a9418ec225ecd5f9a3183697a3dc0dfd3aa1d05d285b3ed42b77c92ae6deee7af6a2b9bc87f119eca9c74810dcc43ddd5eb02b
   google_api_key: <%= ENV["GOOGLE_API_KEY"] %>
   rate_limit_token: pfB6uNKYC8sB9PVgBLdwFToN
+  govuk_basic_auth_credentials: Basic c81a772d7703bfdf8cfc79a0
```
thomasleese: This should not have the `Basic` anymore.

config/application.rb
```diff
@@ -17,6 +17,9 @@
 Bundler.require(*Rails.groups)
 
 module LinkCheckerApi
+  mattr_accessor :authed_domains
+  self.authed_domains = {}
```
thomasleese: Could we rename this to `hosts_with_basic_authorization`.

config/secrets.yml
SebAshton: How do we feel about adding be the `Basic` at initialization? If I was getting a string from a hash called `hosts_with_basic_authorization` I might expect it to have that already.

thomasleese: I would expect a variable called `hosts_with_authorization` to include the `Basic` as it could be any type of authorisation, but since we know they can only be basic authorisation I would expect the strings not to include that.

PR #121, opened 2018-01-15
Restrict urls Google safebrowsing is run on

camillavk:
The usage limit for Google safebrowsing was being exceeded too quickly so
restricting the usage to urls that are not gov.uk ones only.

### DIFF
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -281,7 +281,11 @@ module LinkChecker::UriChecker
     end
 
     def gov_uk_uri?
-      Plek.new.website_root.include?(uri.host)
+      @gov_uk_uri ||= Plek.new.website_root.include?(uri.host)
+    end
+
+    def gov_uk_upload_uri?
+      uri.path.starts_with? "/government/uploads"
     end
 
     def additional_connection_headers
@@ -291,6 +295,8 @@ module LinkChecker::UriChecker
     end
 
     def use_google_safebrowsing?
+      return false if gov_uk_uri? && !gov_uk_upload_uri?
+
       Rails.env.production? || Rails.application.secrets.google_api_key
     end
   end
--- a/spec/lib/link_checker_spec.rb
+++ b/spec/lib/link_checker_spec.rb
@@ -256,16 +256,72 @@ RSpec.describe LinkChecker do
       end
     end
 
-    # context "a URL detected by Google Safebrowser API" do
-    #   let(:uri) { "http://malware.testing.google.test/testing/malware/" }
-    #   before do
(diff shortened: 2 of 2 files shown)


lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -291,6 +291,7 @@ def additional_connection_headers
     end
 
     def use_google_safebrowsing?
+      return false if gov_uk_uri?
```
thomasleese: Mentioned in Slack, but I think there are some URLs we would want to still check such as attachments. Wonder if we can have some sort of whitelist or blacklist.

I do agree with the sentiment though that it's more important that we check external URLs through the Google SafeBrowsing API.

camillavk: @thomasleese I've included a check for if it is a "/government/upload" but this should still reduce the number of requests made significantly

lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -291,6 +295,8 @@ def additional_connection_headers
     end
 
     def use_google_safebrowsing?
+      return false if gov_uk_uri? && !gov_uk_upload_uri?
```
thomasleese: We will probably need to check against URLs like this too:

```
https://assets.publishing.service.gov.uk/media/...
```

camillavk: I had tried to look through a few different links on whitehall to figure out any other urls I would need to check but couldn't really spot any other obvious ones. Do you think the "/government/uploads" and the one you've just mentioned will be the only two needing to be checked?

thomasleese: I think so, yeah. The `/government/uploads` URLs come from the Whitehall attachments system, and the ones I just mentioned come from `asset-manager`.

camillavk: 👍 ok, cool I'll add it in as well then

camillavk: Hmm, @thomasleese I just added an extra test and realised that I don't actually need to modify the code to include this check because the way it is written at the moment it doesn't consider ```https://assets.publishing.service.gov.uk/media/``` to be a `gov_uk_uri` and therefore already sends it for a google safebrowsing request. Do you want me to make it more explicit or clear that it checks these urls or is it ok as it is?

thomasleese: I reckon if we can just add a test to make sure it's not being checked and that should be enough.

camillavk: 👍 added the extra test @thomasleese

PR #122, opened 2018-01-16
Rename GOV.UK Rate Limit configuration

thomasleese:
This makes it clearer that the rate limit is only used for GOV.UK.

Depends on #120.


thomasleese:
Depends on https://github.com/alphagov/govuk-puppet/pull/7093

PR #144, opened 2018-02-26
Disable Google Safebrowsing check

thomasleese:
The API key appears to not be correctly configured and we're seeing access denied errors from the API. For the moment we can disable this check while we figure out how to configure the keys.

### DIFF
--- a/spec/lib/link_checker_spec.rb
+++ b/spec/lib/link_checker_spec.rb
@@ -256,121 +256,126 @@ RSpec.describe LinkChecker do
       end
     end
 
-    context "a URL detected by Google Safebrowser API" do
-      let(:uri) { "http://malware.testing.google.test/testing/malware/" }
-      before do
-        stub_request(:get, uri).to_return(status: 200)
-        stub_request(:post, "https://safebrowsing.googleapis.com/v4/threatMatches:find?key=test")
-          .to_return(status: 200, body: { matches: [{ threatType: "MALWARE" }] }.to_json)
-      end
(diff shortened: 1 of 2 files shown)


spec/lib/link_checker_spec.rb
```diff
@@ -256,121 +256,121 @@
...
-      include_examples "has no errors"
-      it "should increment the counter in statsd" do
-        expect(GovukStatsd).to receive(:increment).with("safebrowsing.rate_limited")
-
-        subject
-
-        expect(request).to have_been_requested
-      end
-    end
+    # context "a URL detected by Google Safebrowser API" do
```
kevindew: Do you want to just add a comment above this to explain why it's commented out?

thomasleese: Added one in now 👍

spec/lib/link_checker_spec.rb
```diff
@@ -256,121 +256,126 @@
...
-      include_examples "has no errors"
-      it "should increment the counter in statsd" do
-        expect(GovukStatsd).to receive(:increment).with("safebrowsing.rate_limited")
-
-        subject
-
-        expect(request).to have_been_requested
-      end
-    end
+    # Commented out as we have disbaled the Google Safebrowsing intergration as the API key
```
kevindew: typo ☝️

thomasleese: Oh yep, thanks!

PR #148, opened 2018-03-13
Use our fork of sidekiq jobs v4.0.18

h-lame:
Sidekiq-unique-jobs 5.x requires redis 3.x but our infrastructure uses
2.8.  We also have to use the fork rather than a released version of 4.x
because the last release of 4.x doesn't include a fix for
mhenrixon/sidekiq-unique-jobs#195 which means the `uniquejobs` hash key
in redis never gets smaller.

Although there is a fix for this in 5.x (see:
https://github.com/mhenrixon/sidekiq-unique-jobs/pulls/200 - this
commit is what is on our fork of 4.x) it may have been changed to rely
on expiry features of redis 3.x that are not available in redis 2.x.

On staging this key is currently 2.5M entries, and consumes ~200Mb.
On production it's only 1.2M entries and consumes ~100Mb.
We tried running a (much simplified version of) this script:
https://gist.github.com/riyad/9086d2b17ff1e8c091cdb1c7ac501b62
in a screen session to remove any expired keys from this hash, but
unfortunately the rate of adding keys to the uniquejobs hash was
greater than the rate of removal.  Instead we waited until the queue
was drained and deleted the key.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,10 @@ gem "faraday-cookie_jar", "~> 0.0.6"
 
 gem "govuk_sidekiq", "~> 3.0"
 gem "sidekiq-scheduler", "~> 2.2"
-gem "sidekiq-unique-jobs", "~> 5.0"
+# We can't use v5 of this because it requires redis 3 and we use 2.8
+# We use our own fork because the latest 4.x release has a bug with
+# removing jobs from the uniquejobs hash in redis
+gem "sidekiq-unique-jobs", git: "https://github.com/alphagov/sidekiq-unique-jobs", branch: 'fix-for-upstream-195-backported-to-4-x-branch', require: false
 
 gem "activerecord-import", "~> 0.22"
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -15,7 +15,10 @@ gem "faraday-cookie_jar", "~> 0.0.6"
 
 gem "govuk_sidekiq", "~> 3.0"
 gem "sidekiq-scheduler", "~> 2.2"
-gem "sidekiq-unique-jobs", "~> 5.0"
+# We can't use v5 of this because it requires redis 3 and we use 2.8
+# We use our own fork because the latest 4.x release has a bug with
+# removing jobs from the uniquejobs hash in redis
+gem "sidekiq-unique-jobs", git: "https://github.com/alphagov/sidekiq-unique-jobs", require: false
```
suzannehamilton: Is it worth tagging the current master of alphagov/sidekiq-unique-jobs and referencing it here? Unless you think that lock file will be enough to stop people from upgrading this dependency without checking it? (In the scenario that someone happens to update our fork of the gem because they don't know what it's being used for, or they want to use it in some other way.)

h-lame: Oh, yes. That's a v. good idea. I'll do that and change this PR (and change the publishing-api gemfile again because I already merged it).

h-lame: Created a branch `fix-for-upstream-195-backported-to-4-x-branch` and specified that.  I figure a branch lets us roll it forward more easily rather than a tag in case we find problems with the fix.

PR #174, opened 2018-05-03
Improve SQL query to remove old batches and checks

tijmenb:
The query generated in this worker removes old batches and checks. It does so by `plucking` all the IDs of the old batches, and running a delete query on them.

When the number of old batches exceeds 65535, this will cause an SQL error:

```
ActiveRecord::StatementInvalid: PG::UnableToSend: number of parameters must be between 0 and 65535\n: DELETE FROM \"batches\" WHERE \"batches\".\"id\" IN ($1, $2, $3, $4, $...
```

This in turn causes billions of log files from Sidekiq, which is currently taking down machines on integration.

This commit changes the query so that we feed in the `BatchCheck.where(check: [old_checks])` directly, which Rails will magically use to generate a subquery, as evidenced by running this in the console:

```rb
irb(main):008:0> puts Batch.where(id: BatchCheck.where(check: [old_checks])).to_sql
SELECT "batches".* FROM "batches" WHERE "batches"."id" IN (SELECT "batch_checks"."id" FROM "batch_checks" WHERE "batch_checks"."check_id" IN (SELECT "checks"."id" FROM "checks" WHERE (completed_at < '2018-04-05 10:06:49.607593')))
```

We've improved the tests a bit to make sure that we haven't broken anything.

### DIFF
--- a/app/workers/cleanup_worker.rb
+++ b/app/workers/cleanup_worker.rb
@@ -25,6 +25,6 @@ private
   end
 
   def old_batches
-    Batch.where(id: [BatchCheck.where(check: [old_checks]).pluck(:batch_id)])
+    Batch.where(id: BatchCheck.select(:batch_id).where(check: old_checks))
   end
 end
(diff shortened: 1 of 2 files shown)


bevanloon:
@thomasleese tweaked the query slightly - can you take another quick look please?

PR #176, opened 2018-05-08
Update README

tijmenb:
- Remove "alpha" things and outdated history
- Add emoji
- Remove API output (it's already in the docs)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,7 @@
 # Link Checker API
 
-**⚠️ This service is in Alpha ⚠️**
-
 A web service that takes an input of URIs. It performs a number of checks on
 them to determine whether these are things that should be linked to.
-
-This service was created in the GOV.UK 2 week blitz of April 2017, thoughts on
-next steps are documented in [docs/blitz.md](docs/blitz.md).
 
 ## Nomenclature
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,8 @@
 # Link Checker API
 
-**⚠️ This service is in Alpha ⚠️**
-
 A web service that takes an input of URIs. It performs a number of checks on
 them to determine whether these are things that should be linked to.
 
-This service was created in the GOV.UK 2 week blitz of April 2017, thoughts on
```
thomasleese: I would prefer not to remove this as it's still very relevant if a new team were to pick up this work.

tijmenb: I'm not sure, it's very unclear to me what the status of that document is. How can there have been a 3 month project to improve the link checker and no updates to that doc? Do you think it's still relevant?

I wonder if it should be removed and added to a Trello board, I don't think any product manager who would pick this up would think of reading the README/docs folder.

thomasleese: The only thing that can be removed from that document is the section about integration with Whitehall, as that (and integration with all the other publishing apps) was completed in the 3 month project.

I definitely agree that these would be better placed on a Trello board though...

PR #331, opened 2020-01-22
Treat SSL issues as warnings

barrucadu:
There are many local authority websites which have misconfigured SSL
certs: they only send the leaf cert and not the full chain.  This is
incorrect according to the standard, and SSL checker tools (like
ssllabs.com) will flag this up.  But browsers will often work with
such websites: they cache certs, and so if the user has visited a
website which does serve the intermediate cert, then the website which
doesn't will work.

Since a user visiting the page in a browser gets the page, the link
"works", even though it technically has a problem.  There are a lot of
websites with this issue, so it really diminishes the value of the
link-checker-api to report these as hard errors.

We're not in the business of monitoring other people's websites for
configuration issues, so we're downgrading this issue to a warning.

---

[Trello card](https://trello.com/c/pKNcv0og/1279-investigate-link-checker-api-tls-failures)

### DIFF
--- a/lib/link_checker/uri_checker/http_checker.rb
+++ b/lib/link_checker/uri_checker/http_checker.rb
@@ -83,6 +83,12 @@ module LinkChecker::UriChecker
     end
   end
 
+  class SecurityProblem < LinkChecker::UriChecker::Warning
+    def initialize(options = {})
+      super(summary: :security_problem, message: :page_has_security_problem, suggested_fix: :contact_site_administrator, **options)
+    end
+  end
+
   class FaradayError < Error
     def initialize(options = {})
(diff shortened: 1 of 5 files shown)


lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -249,14 +255,18 @@ def make_request(method)
         )
         nil
       rescue Faraday::SSLError
+        # if we've ended up here again, just abort
+        return nil unless check_ssl
```
barrucadu: This will only be reachable if we have a bug, so I'm open to a less-silent way of handling this issue.

PR #345, opened 2020-02-27
Use `rubocop-govuk`, not `govuk-lint`

issyl0:
- The `govuk-lint` gem is deprecated. The new standard is to run RuboCop
  with an inherited set of rules.
- Only lint standard Ruby language problems, don't lint Rails as well as
  there are way more errors. That can be a stretch goal.


issyl0:
There's a load of deprecation warnings in the CI output for code/gem things, but I'll leave those for someone else.

PR #352, opened 2020-03-10
Make entire repository pass Rubocop linting

ChrisBAshton:
This is depended on by https://github.com/alphagov/govuk-jenkinslib/pull/66.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,3 +5,7 @@ inherit_gem:
 inherit_mode:
   merge:
     - Exclude
+
+Metrics/BlockLength:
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -5,3 +5,7 @@ inherit_gem:
 inherit_mode:
   merge:
     - Exclude
+
+Metrics/BlockLength:
+  Exclude:
+    - db/schema.rb
```
issyl0: If you also enable Rails cops, `db/schema.rb` will be excluded automatically. But that's probably way more work.

PR #406, opened 2020-08-05
Make app consistent with upgrades

cdccollins:
This makes Link Checker API consistent with the other apps upgraded to Rails 6


app/models/check.rb
```diff
@@ -7,7 +7,7 @@ class Check < ApplicationRecord
   belongs_to :link
 
   scope :created_within, ->(within) { where("created_at > ?", Time.zone.now - within) }
-  scope :requires_checking, -> { where(started_at: nil).or(Check.where(completed_at: nil).where("created_at < ?", RECHECK_THRESHOLD.ago)) }
+  scope :requires_checking, -> { where(started_at: nil).or(self.where(completed_at: nil).where("created_at < ?", RECHECK_THRESHOLD.ago)) }
```
Rhian, line 10: Rubocop suggests removing the `self`:
`scope :requires_checking, -> { where(started_at: nil).or(where(completed_at: nil).where("created_at < ?", RECHECK_THRESHOLD.ago)) }`

PR #411, opened 2020-09-16
Always print code coverage after running tests

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

We should always monitor the amount of code that is exercised when
we run our tests.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,9 +1,6 @@
-if ENV["RCOV"]
-  require "simplecov"
-  require "simplecov-rcov"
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -1,9 +1,6 @@
-if ENV["RCOV"]
-  require "simplecov"
-  require "simplecov-rcov"
```
kevindew, line 3: Can we also drop this from the Gemfile?

benthorner: Oops, missed that this time round - updated 👍

kevindew (approved):
Nice one

PR #436, opened 2021-01-08
Support technically invalid, but still working, URIs

thomasleese:
Often there will be URIs which are technically invalid (for example using the `[` character instead of percent encoding it) but browsers will still accept them to, and do the encoding itself before making the request.

This PR updates Link Checker API to mimic the behaviour of browsers so that we see fewer invalid URIs which would actually work in the real world.

An example URI which was flagged as invalid was `https://m.luton.gov.uk/Page/Show/Council_tax/help-[paying-or-reducing-your-council-tax/Pages/default.aspx?redirectToMobile=True` which does work if you go to it in a browser.

[Trello Card](https://trello.com/c/YSVcZtIB/2297-3-link-checking-reports-a-working-url-as-invalid)

### DIFF
--- a/app/lib/link_checker/uri_checker/http_checker.rb
+++ b/app/lib/link_checker/uri_checker/http_checker.rb
@@ -279,7 +279,7 @@ module LinkChecker::UriChecker
 
     def run_connection_request(method, check_ssl: true)
       client = check_ssl ? http_client : insecure_http_client
-      client.run_request(method, uri, nil, additional_connection_headers) do |request|
+      client.run_request(method, uri.normalize, nil, additional_connection_headers) do |request|
         request.options[:timeout] = RESPONSE_TIME_LIMIT
         request.options[:open_timeout] = RESPONSE_TIME_LIMIT
       end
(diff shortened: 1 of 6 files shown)


app/lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -97,7 +97,7 @@ def initialize(options = {})
 
   class HttpChecker < Checker
     def call
-      if uri.host.nil?
+      if uri.host.blank?
```
ChrisBAshton, line 100: Should there be corresponding tests for this change?

ChrisBAshton (approved):
LGTM 👍 Easy to follow commits. Thanks!

app/lib/link_checker/uri_checker/http_checker.rb
thomasleese: There kind of already is in the sense that the tests start failing after https://github.com/alphagov/link-checker-api/pull/436/commits/5e095ca0bfd9aae18739ae923a83baa6eebd0462 but work again after https://github.com/alphagov/link-checker-api/pull/436/commits/bd4ca57ca6b0a74b5ad83d09f157f66344fe8221. Do you think that's enough?

ChrisBAshton: Ah, I missed that context. Yes that seems fine! 👍

PR #440, opened 2021-01-27
Set User-Agent to a recent version of Chrome

thomasleese:
It appears that recently Twitter has started checking the User-Agent to decide whether to serve a response to the user.

```
❯ curl -I https://twitter.com/DefraStats
HTTP/2 400
...
<p>
  Please switch to a supported browser to continue using twitter.com. You can see a list of supported browsers in our Help Center.
</p>
```

```
❯ curl -I https://twitter.com/DefraStats -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36'
HTTP/2 200
```

By pretending to be a reasonably common Chrome browser, we should avoid any potential issues with servers blocking our requests. This change came from a Zendesk ticket where a user was seeing their twitter links marked as broken: https://govuk.zendesk.com/agent/tickets/4442955

### DIFF
--- a/app/lib/link_checker/uri_checker/checker.rb
+++ b/app/lib/link_checker/uri_checker/checker.rb
@@ -30,8 +30,10 @@ module LinkChecker::UriChecker
 
     attr_reader :redirect_history
 
+    USER_AGENT = "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36".freeze
+
     def client(options = {})
-      default_options = { headers: { accept_encoding: "none" } }
+      default_options = { headers: { accept_encoding: "none", user_agent: USER_AGENT } }
       Faraday.new(default_options.merge(options)) do |faraday|
         faraday.use :cookie_jar
         faraday.adapter Faraday.default_adapter


1pretz1 (approved):
Love a bit of spoofing! 😄

kevindew:
Oh man this seems a pain, nice job working it out. It is a shame it's only needed for Twitter. I guess the code isn't really set up for it to be conditional based on host. It does seem a bit annoying that the version of Chrome we use in the User Agent will fall out of support and a 400 will start occurring - and we'll have no idea of it.

For example:

```
➜  ~ curl -sI https://twitter.com/DefraStats -H 'User-Agent: Chrome/37.0.0.0' | grep HTTP
HTTP/2 400
➜  ~ curl -sI https://twitter.com/DefraStats -H 'User-Agent: Chrome/38.0.0.0' | grep HTTP
HTTP/2 200
```

Without going down a more complicated path though I wonder if this would be better with a tailored user agent header with the spoof for example: 

```
➜  ~ curl -sI https://twitter.com/DefraStats -H 'User-Agent: GOV.UK Link Checker - Chrome/80.0.0.0' | grep HTTP
HTTP/2 200
➜  ~ curl -sI https://twitter.com/DefraStats -H 'User-Agent: GOV.UK Link Checker - Chrome/1000.0.0.0' | grep HTTP
HTTP/2 200
```

thomasleese:
@kevindew Yeah, I did think about doing something like that, but I kind of figured that whatever we put here has the chance of breaking in the future.

I thought by using a `User-Agent` that is a known browser string we'd have more chance of success whereas if we created our own (even though it's more accurate) it's more likely to be blocked again.

kevindew:
> @kevindew Yeah, I did think about doing something like that, but I kind of figured that whatever we put here has the chance of breaking in the future.
> 
> I thought by using a `User-Agent` that is a known browser string we'd have more chance of success whereas if we created our own (even though it's more accurate) it's more likely to be blocked again.

Sure - absolutely everything has a chance of breaking in future, super annoying thing Twitter has done :( The real bummer of this being totally a browser spoof is no-one can ever identify if we're doing something nasty. It'd be good if it had some references to this tool?

thomasleese:
@kevindew I've opened https://github.com/alphagov/link-checker-api/pull/441, what do you think?

PR #441, opened 2021-01-29
Update user agent

thomasleese:
Follow up for #440.

### DIFF
--- a/app/lib/link_checker/uri_checker/checker.rb
+++ b/app/lib/link_checker/uri_checker/checker.rb
@@ -30,7 +30,7 @@ module LinkChecker::UriChecker
 
     attr_reader :redirect_history
 
-    USER_AGENT = "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36".freeze
+    USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36".freeze
 
     def client(options = {})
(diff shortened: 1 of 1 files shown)


app/lib/link_checker/uri_checker/checker.rb
```diff
@@ -30,7 +30,7 @@ def insecure_http_client
 
     attr_reader :redirect_history
 
-    USER_AGENT = "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36".freeze
```
kevindew, line 33: Oops! I hadn't noticed this - good catch.

app/lib/link_checker/uri_checker/checker.rb
```diff
@@ -30,7 +30,7 @@ def insecure_http_client
 
     attr_reader :redirect_history
 
-    USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36".freeze
+    USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 (GOV.UK Link Checker API)".freeze
```
kevindew, line 33: This looks a good improvement 👍  - still think it should be less spoofy but not a hill I'm going to die on.

Shall we update Chrome to be the current version, [58 is from 2017](https://developer.chrome.com/blog/new-in-chrome-58/)

kevindew (approved):

app/lib/link_checker/uri_checker/checker.rb
thomasleese: Good point, I've push up a commit which fixes that.

PR #446, opened 2021-03-03
Return early from check_request if response is nil

benjamineskola:
`make_request` returns nil if Faraday raises an exception. These are already added to the report by `make_request`, so there's no need to do so again here. The return value of `check_request` is not used.

https://trello.com/c/CHUwfu4h/2390-3-fix-nomethoderror-in-link-checker-api

### DIFF
--- a/app/lib/link_checker/uri_checker/http_checker.rb
+++ b/app/lib/link_checker/uri_checker/http_checker.rb
@@ -154,7 +154,7 @@ module LinkChecker::UriChecker
       end_time = Time.zone.now
       response_time = end_time - start_time
 
-      return response if report.has_errors?
+      return response if report.has_errors? || response.nil?
 
       add_problem(SlowResponse.new(from_redirect: from_redirect?)) if response_time > RESPONSE_TIME_WARNING
 
--- a/spec/lib/http_checker_spec.rb
+++ b/spec/lib/http_checker_spec.rb
@@ -1,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe LinkChecker::UriChecker::HttpChecker do
+  let(:uri) { URI("http://example.invalid/") }
+  subject { described_class.new(uri) }
+  it "should not throw an exception if the response is nil" do
+    allow(subject).to receive(:make_request).with(:get).and_return(nil)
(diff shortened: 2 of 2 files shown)


app/lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -158,6 +158,8 @@ def check_request
 
       add_problem(SlowResponse.new(from_redirect: from_redirect?)) if response_time > RESPONSE_TIME_WARNING
 
+      return if response.nil?
```
ChrisBAshton, line 161: Looking good so far!

I'd be tempted to move this onto line 157, to minimise the number of places the method can return from:

```ruby
      return response if report.has_errors? || response.nil?
```

Both have the effect of returning `nil`, though as you've pointed out, the return value isn't used.

Also, is it possible to write a test for this?

spec/lib/http_checker_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe LinkChecker::UriChecker::HttpChecker do
+  let(:uri) { URI("http://example.invalid/") }
+  subject { described_class.new(uri) }
+  it "should not throw an exception if the response is nil" do
+    stub_request(:get, "http://example.invalid/").to_return(status: 500, body: "", headers: {})
```
ChrisBAshton, line 7: We can DRY this up - and I also don't think we necessarily need an empty body or headers 🤔 

```suggestion
    stub_request(:get, uri).to_return(status: 500)
```

I also wonder if this `stub_request` is actually doing anything - I imagine the `allow(subject).to receive` on the next line is setting up the test how we want. What happens if we remove `stub_request`?

ChrisBAshton (approved):
LGTM, just a minor point on the test stubbing.

spec/lib/http_checker_spec.rb
benjamineskola: Ah, I added the `stub_request` because of an error but that seems like it was a red herring — presumably I'd gotten the other stub wrong at that stage.

ChrisBAshton (approved):
Great! 👏

PR #455, opened 2021-04-09
Decouple Raven from Link-checker-api

callumknights:
Removed direct references to the Raven object and renamed appropriately

Trello: https://trello.com/c/S1QvqyCE/2459-decouple-govuk-apps-from-raven-3

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,4 +1,4 @@
-Raven.configure do |config|
+GovukError.configure do |config|
   config.excluded_exceptions += %w[
     RestartWorkerException
   ]


db/schema.rb
```diff
@@ -2,24 +2,25 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
+# This file is the source Rails uses to define your schema when running `rails
```
ChrisBAshton, line 5: These changes look like they've crept in. Not the end of the world, but they should be in a separate commit explaining where they've come from (or just omitted from this PR).

ChrisBAshton (commented):
The main change looks good, but one comment below:

db/schema.rb
callumknights: Ahh, this is weird, I think this is a result of a change the linter demanded of me, I'm not entirely sure why it's also applied to the db schema

ChrisBAshton (approved):

PR #491, opened 2021-12-01
Clean database after Pact tests

ollietreend:
This commit changes the configuration of the Pact tests.

They were previously configured to truncate the database before each Pact example. However, this meant that the final example in the Pact test suite wouldn't clean up after itself, causing records to be left in the database.

This resulted in subsequent test runs to fail because the test suite expected to begin with an empty database.

This change configures Pact to use the 'transaction' strategy of Database Cleaner to run each example in its own database transaction. The database transaction is started before each Pact test, and is rolled back after each test finishes.

The result is that the database remains clean after a full run of the test suite.

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -31,8 +31,13 @@ end
 
 Pact.provider_states_for "GDS API Adapters" do
   set_up do
-    DatabaseCleaner.clean_with :truncation
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.start
     GDS::SSO.test_user = create(:user, permissions: %w[signin])
+  end
+
+  tear_down do
+    DatabaseCleaner.clean
   end
 
   provider_state "a batch exists with id 99 and uris https://www.gov.uk" do


kevindew (commented):
This looks good to me - nice one. Do we definitely need the `clean` in the `tear_down`? With a transaction strategy I guess it's just never commited.

In general we've moved away from using DatabaseCleaner since Rails included the ability to run the tests in transactions (`use_transactional_tests`) - which is normally applied to RSpec with something like: https://github.com/alphagov/content-publisher/blob/main/spec/spec_helper.rb#L33

I wonder if we'd be able to do the same here by adding something like: `config.use_transactional_fixtures = true` to: https://github.com/alphagov/link-checker-api/blob/c5ac81b5dd3a0c116b5bc023971b29f4adfe59af/spec/service_consumers/pact_helper.rb#L5 ? 

as per: https://github.com/alphagov/link-checker-api/blob/8243198bce790ca356445ec0c767f479ccf8a0db/spec/rails_helper.rb#L19

and then get rid of the database_cleaner gem? I suggest this with low idea of it will work

kevindew:
Hmm with some digging it doesn't look so possible, which is a shame: https://github.com/pact-foundation/pact-ruby/issues/167 - I'll approve this.

kevindew (approved):

ollietreend:
Thanks @kevindew. It just so happened that I stumbled across [that same issue](pact-foundation/pact-ruby#167), which led me to the same solution. 😄

PR #557, opened 2022-09-23
Handle argument errors when http request raises argument error for headers having CR/LF characters

BeckaL:
Faraday / ruby cannot parse headers which contain CR/LF characters. As this issue appears to exist deep down in the faraday / ruby stack, we should handle the error rather than letting the application alert. This change attempts to handle only those argument errors that roughly fuzzy match the exception string 'header field value cannot include CR/LF' (with a bit of fuzziness in case the exact message changes) but still raise for other exceptions.

[Trello](https://trello.com/c/eZ70YojB/291-fix-for-sentry-errors-header-field-value-cannot-include-cr-lf)

### DIFF
--- a/app/lib/link_checker/uri_checker/http_checker.rb
+++ b/app/lib/link_checker/uri_checker/http_checker.rb
@@ -89,7 +89,7 @@ module LinkChecker::UriChecker
     end
   end
 
-  class FaradayError < Error
+  class HttpCommunicationError < Error
     def initialize(options = {})
       super(suggested_fix: :determine_if_temporary, **options)
     end
@@ -238,7 +238,7 @@ module LinkChecker::UriChecker
       response
     rescue Faraday::ConnectionFailed
       add_problem(
-        FaradayError.new(
+        HttpCommunicationError.new(
           summary: :website_unavailable,
           message: :website_host_offline,
           from_redirect: from_redirect?,
@@ -247,7 +247,7 @@ module LinkChecker::UriChecker
       nil
     rescue Faraday::TimeoutError
       add_problem(
-        FaradayError.new(
+        HttpCommunicationError.new(
           summary: :website_unavailable,
           message: :page_is_not_responding,
           from_redirect: from_redirect?,
@@ -269,9 +269,26 @@ module LinkChecker::UriChecker
       make_request(method, check_ssl: false)
     rescue Faraday::Error
       add_problem(
-        FaradayError.new(
+        HttpCommunicationError.new(
           summary: :page_unavailable,
-          message: :page_failing_to_load,
+          message: :technical_error_on_page,
+          from_redirect: from_redirect?,
+        ),
+      )
+      nil
+    # Ruby net-http cannot handle responses with headers with CR/LF characters in, and such responses raise
+    # an argument error. We want to catch these and add to the report for now, rather than continuing to raise these
+    # as errors, which can trigger alerting. We are planning on raising this to be fixed at the net-http level which may
+    # mean we do not have to handle these here.
+    # Doing some fuzzy matching on the error message to try and ensure a bit of resilience, allowing for changes to the
+    # exact error message in the exception.
+    rescue ArgumentError => e
+      raise e unless e.message =~ /(.*)header(.*) CR\/LF/
+
+      add_problem(
+        HttpCommunicationError.new(
+          summary: :page_unavailable,
+          message: :technical_error_on_page,
           from_redirect: from_redirect?,
         ),
       )
--- a/spec/lib/link_checker_spec.rb
+++ b/spec/lib/link_checker_spec.rb
@@ -142,6 +142,26 @@ RSpec.describe LinkChecker do
       include_examples "has warnings"
     end
 
+    context "header with CR/LF character" do
+      let(:uri) { "http://www.not-gov.uk/header_with_CRLF_character" }
+      before do
+        stub_request(:get, uri)
+          .to_return(headers: { "Invalid" => "A header containing a carriage return \r character" })
+      end
+
+      include_examples "has errors"
+      include_examples "has a problem summary", "Page unavailable"
+    end
+
+    it "does not recue from other argument error" do
+      uri = "http://www.not-gov.uk/raises_argument_error"
+      error = ArgumentError.new("something that's nothing to do with headers and carriage return line feed chars")
+
+      stub_request(:get, uri).to_raise(error)
+
+      expect { described_class.new(uri).call }.to raise_error(error)
+    end
+
     context "slow response" do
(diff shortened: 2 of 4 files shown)


app/lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -230,12 +236,24 @@ def make_request(method, check_ssl: true)
...
+      case e.message
+      when /header (.*) CR\/LF/
+        add_problem(
+          CRLFArgumentError.new(
+            from_redirect: from_redirect?,
+          ),
+        )
+        nil
+      else
+        raise e
```
BeckaL, line 255: Is there a cleaner way of doing this, by just rescuing from Argument Errors matching the specific string, and not having to specify the case where the message string doesn't match and having to raise again?

spec/lib/link_checker_spec.rb
```diff
@@ -142,6 +142,26 @@
       include_examples "has warnings"
     end
 
+    context "header with CR/LF character" do
```
BeckaL, line 145: going to make this description a bit better

spec/lib/link_checker_spec.rb
```diff
@@ -285,7 +305,7 @@
         subject
 
         expect(WebMock).to have_requested(:get, uri)
-          .with(headers: { "Rate-Limit-Token": Rails.application.secrets.govuk_rate_limit_token, "Accept-Encoding": "none" })
+                             .with(headers: { "Rate-Limit-Token": Rails.application.secrets.govuk_rate_limit_token, "Accept-Encoding": "none" })
```
BeckaL, line 308: will revert these formatting changes, running rubocop locally added them in 🤔

app/lib/link_checker/uri_checker/http_checker.rb
```diff
@@ -236,6 +242,18 @@ def make_request(method, check_ssl: true)
       end
 
       response
+    rescue ArgumentError => e
```
kevindew, line 245: We should have this as the last exception we rescue from since it's the most generic. This would mean that if any of the earlier exceptions inherit from ArgumentError they'd still be handled first. 

As it's pretty weird we should also put a comment above it to stay that it's an error type that Faraday doesn't handle.

app/lib/link_checker/uri_checker/http_checker.rb
kevindew: We can't use Ruby's pattern matching here so this is the most common approach. Though it'd probably be a little cleaner with

```ruby
raise e unless e.message ~= /header (.*) CR\/LF/

add_problem(...
```

Though Googling it shows you can apply some clever approaches: https://www.honeybadger.io/blog/level-up-ruby-rescue-with-dynamic-exception-matchers/ but they might be a bit cryptic.

spec/lib/link_checker_spec.rb
```diff
@@ -142,6 +142,26 @@
       include_examples "has warnings"
     end
 
+    context "header with CR/LF character" do
+      let(:uri) { "http://www.not-gov.uk/header_with_CRLF_character" }
+      before do
+        stub_request(:get, uri)
+          .to_raise(ArgumentError.new("header field value cannot include CR/LF"))
```
kevindew, line 149: I think we can stub this in a somewhat more realistic way with:

```
        stub_request(:get, uri)
          .to_return(headers: { "Invalid" => "With a carriage return \r character" })
```

This could be a bit more useful as we'd learn if a dependency change broke the fix.

kevindew (commented):
Nice work putting this together, it looks like it'll do the trick. I've made some suggestions to try make this a bit more of a generic implementation. Happy to chat if they don't make sense. 

I'll put on my todo list to raise an issue with Ruby net-http to see if it can be fixed at the library.

BeckaL:
@kevindew thanks for the helpful comments! Should all be addressed now

kevindew:
Thanks @BeckaL that looks good, thanks for addressing my comments. 

I managed to fail to send a comment somehow too, sorry - it was probably my longest comment as well 🤦 

I think we should have a more generic exception than `CRLFArgumentError` as I think that is too specific for the users of this link checker as it's very technical. Since we would have expected that this error would be caught with the [generic Faraday error](https://github.com/alphagov/link-checker-api/blob/2afe837ec9d684bab10a7f323aa6b33763ccb710/app/lib/link_checker/uri_checker/http_checker.rb#L276-L284) I reckon we can use the same messaging as that.

I also wondered if rather than needing to create a new error exception we could repurpose the FaradayError one. I think we've created that as basically the generic failure error. I wonder if we could just call it something like CommunicationError and use it rather than creating a new user facing error for this scenario.

Thanks again, and sorry about the late comment.

BeckaL:
Yup good point @kevindew I was struggling to work out good messaging and naming, will have a look 🙂

BeckaL:
Hey @kevindew - I've pushed my suggestion for this, a lot of it from what you suggested. I have:

* Removed CRLFArgumentError
* Renamed FaradayError to HttpCommunicationError (as this one is just in code rather than manifested externally, think it makes it a bit more specific for developers than just CommunicationError);
* Changed the error _summary_ in the case of hitting the CR/LF case to be the generic 'page unavailable' error
* Still kept an error _message_ that's specific to this case. 

My reasoning for this last one is that it's a bit of a balancing act - the message is definitely a bit technical for the layperson. However, I'm thinking about guarding against a more generic 'page couldn't load' error coming to someone in the report, where they'd clearly be able to see that the page does load in their browser, so worried about people losing faith in the link checker with this more generic message. I'm also thinking that it'll be easier for a developer to quickly diagnose if the error message is specific, so thinking about the diagnosis effort for future zendesk tickets. Have attempted to word it in the least technical way I can, but I guess as soon as you have to talk about headers, it's tricky.

What do you reckon?

kevindew:
Thanks @BeckaL looks great.

Your argument makes sense, I can see where you're coming from. My perspective is that, were Ruby's net-http returning a different exception that Faraday was catching, we would not be putting any additional messaging for this and they'd get the generic pairing of "Page unavailable", "This page is failing to load." and our reasons for not using that error here probably also apply to a generic faraday error.

I think mentioning anything about headers is going too technical compared to others, if we were to do it I'd probably go down an ambiguous "technical misconfigured". I'm not worried about the debugging situation as it only distinguishes it from one of multiple sources of the above generic error.

We could change the generic error to be something like "This page has a technical problem"

BeckaL:
OK yep I follow your reasoning, that makes sense. Updated the generic error as you suggested to be 'This page has a technical problem'.

kevindew (approved):
Great stuff

kevindew:
I've opened: https://github.com/ruby/net-http/pull/72 to try resolve this upstream

PR #559, opened 2022-09-29
Increase test coverage

brucebolt:
Before enabling continuous deployment for this application, we need to increase the test coverage.

This PR increases coverage from 93.58% to 97.28% by removing code that never gets executed and adding additional scenarios where they were not previously being tested.

Additionally, it suppresses some logging to the console that was spamming the output of the tests.

[Trello card](https://trello.com/c/xyWyqqhq/15-enable-continuous-deployment-for-link-checker-api)


BeckaL (approved):
All looks good to me, nice cleanup to remove that unused safebrowsing code, new tests look good, and so good to have fixed that annoying test logging, will make reading the test output so much nicer! And >97% test coverage ⭐

PR #561, opened 2022-09-29
Upgrade Ruby to 3.1.2

brucebolt:
Upgrades Ruby to 3.1.2. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.

[Trello card](https://trello.com/c/Ab0lArmp)


BeckaL (approved):
All looks sensible to me, will also apply the bundler updates to mine

PR #733, opened 2023-10-31
Upgrade to Rails 7.1

brucebolt:
Rails 7.1.1 was updated in https://github.com/alphagov/link-checker-api/pull/725 without also running the Rails update task.
    
This runs the task, updating the config where necessary.

[Trello card](https://trello.com/c/cNcPQMVF)


config/environments/production.rb
```diff
@@ -62,50 +69,11 @@
...
-  config.active_support.disallowed_deprecation_warnings = []
-
-  # Use default logging formatter so that PID and timestamp are not suppressed.
-  config.log_formatter = ::Logger::Formatter.new
-
-  # Use a different logger for distributed setups.
-  # require "syslog/logger"
-  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
-
-  if ENV["RAILS_LOG_TO_STDOUT"].present?
```
jkempster34, line 81: Does it matter that logging to stdout is now the default instead of being controller by this env variable?

config/environments/test.rb
```diff
@@ -28,7 +29,7 @@
   config.cache_store = :null_store
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  config.action_dispatch.show_exceptions = :rescuable
```
jkempster34, line 32: Just read up on this, seems `:rescuable` is more appropriate than `:none` 👍

jkempster34 (approved):
LGTM. One question about logging

config/environments/production.rb
brucebolt: Good spot.  I think I'll retain that environment variable to maintain the existing behaviour.

PR #750, opened 2023-11-29
Check suspicious domains

richardTowers:


KludgeKML (approved):
LGTM

richardTowers:
Bruce pointed out that the link-checker doesn't automatically scan all the links (as I assumed it did), but there's [a rake task](https://github.com/alphagov/whitehall/blob/main/lib/tasks/broken_link_reporting.rake) in Whitehall to make it scan things by org.

I think I'll still merge this, and see if the rake task flags anything up.

PR #811, opened 2024-04-02
Do not raise when a webhook worker experiences a remote server error

brucebolt:
This application returns the link checking results to the requesting application through a webhook.

In some cases, the remote server will respond with an error (e.g. a timeout). We don't need to log these to Sentry, since they are outside of this application's control.

Therefore rescuing from the `Faraday::ServerError` raise that can occur in this worker.

[Trello card](https://trello.com/c/VjWY3KPa)

### DIFF
--- a/app/workers/webhook_worker.rb
+++ b/app/workers/webhook_worker.rb
@@ -34,6 +34,8 @@ class WebhookWorker
   rescue Faraday::ClientError => e
     logger.error e.message
     raise RestartWorkerException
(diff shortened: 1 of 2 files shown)


JonathanHallam (approved):
nice one, thanks!

brucebolt:
Thanks @JonathanHallam.  Would you mind looking again, since I forgot to commit the test 🤦‍♂️!

PR #969, opened 2025-03-03
Introduce 'Danger' response for suspicious domains

JonathanHallam:
The publishing teams have asked for this change. It will allow the api to respond with "danger" when a link is suspicious which means the teams can separate out these links to disallow them completely. 

Will need supporting changes to the pact tests in gds api adapters

### DIFF
--- a/app/lib/link_checker/uri_checker/danger.rb
+++ b/app/lib/link_checker/uri_checker/danger.rb
@@ -1,0 +1,7 @@
+module LinkChecker::UriChecker
+  class Danger < Problem
+    def initialize(**options)
+      super(type: :danger, **options)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


brucebolt (commented):
Do the `danger` values need adding to the [`CheckPresenter`](https://github.com/alphagov/link-checker-api/blob/1243d770778ab3a53b6e72be1393414f9e4451bf/app/presenters/check_presenter.rb#L8-L9), so they are output to the requesting app?

There's also [a `status` value](https://github.com/alphagov/link-checker-api/blob/1243d770778ab3a53b6e72be1393414f9e4451bf/app/models/check.rb#L50-L56) that looks like it'll output `ok` if there are `danger` links present.

brucebolt (approved):

